PR #1, opened 2015-04-17
Render signup pages

tommyp:
In order to allow users to sign up to alerts from govdelivery, EmailAlertFrontend needs to render signup pages which provide a link to a govdelivery URL which is generated from a content item. The details part of the content item is formatted like:

```
  "details": {
    "breadcrumbs": [
      {
        "title": "Employment",
        "link": "https://www.gov.uk/government/policies/employment"
      }
    ],
    "tags": {
      "policy": [
        "employment"
      ]
    }
  },
```

`breadcrumbs` is an array of parent links and `tags` is a hash with each item in the hash being an array of the tag values.

This PR also does a bunch of setup things such as adding `govuk` gems, `jenkins.sh` and `jenkins_branches.sh` along with setting up RSpec and Cucumber.

Screenshot:

screen shot 2015-04-17 at 16 50 08

Requires https://github.com/alphagov/govuk-content-schemas/pull/42 to be merged before tests will pass.

- [x] merge https://github.com/alphagov/govuk-content-schemas/pull/42


app/models/email_alert_signup.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def find_or_create_subscription
+    EmailAlertFrontend.services(:email_alert_api)
+      .find_or_create_subscriber_list(subscription_params)
+  end
+
+  def breadcrumbs
+    return {} if raw_breadcrumbs.blank?
+
+    raw_breadcrumbs.reverse.reduce { |memo, crumb|
```
tekin: :+1:

edds:
I went through the HTML/Sass changes with @tommyp and think they are good.

features/support/email_alert_signup_helper.rb
```diff
@@ -0,0 +1,26 @@
+require 'ostruct'
+
+module EmailAlertSignupHelper
+  include GovukContentSchemaExamples
+
+  def content_store_has_employment_email_alert_signup(base_path:, tags:)
+    content_store_has_item(base_path, govuk_content_schema_example("email_alert_signup").merge(tags).to_json)
+  end
+
+  def expect_registration_to(title:, tags:, base_path:)
```
tekin: This expectation appears to be setting a stub rather than an actual expectation. Does that mean the scenario will always pass, even if the subscription never happens?

tekin: I think what is missing is an additional expectation:

```ruby
    expect(EmailAlertFrontend.services(:email_alert_api)).
      to have_received(:find_or_create_subscriber_list).
      with("title" => title,"tags" => tags)
```

lib/govuk_content_schema_examples.rb
```diff
@@ -0,0 +1,29 @@
...
+  config.project_root = Rails.root
+end
+
+module GovukContentSchemaExamples
+  extend ActiveSupport::Concern
+
+  included do
+    include GdsApi::TestHelpers::ContentStore
+
+    # Returns a hash representing an finder content item from govuk-content-schemas
```
tekin: "... representing an *email alert* signup content item"

spec/models/email_alert_signup_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+    )
+  }
+
+  let (:email_alert_api) { double(:email_alert_api) }
+
+  before do
+
+    EmailAlertFrontend.register_service(:email_alert_api, email_alert_api)
+
+    allow(email_alert_api).to receive(:find_or_create_subscriber_list).and_return(OpenStruct.new("subscriber_list" => OpenStruct.new("subscription_url" => "http://govdelivery_signup_url")))
```
tekin: Would these specs be less brittle if they used the pre-existing helpers that ship with `gds-api-adapters`? e.g. [GdsApi::TestHelpers::EmailAlertApi#email_alert_api_creates_subscriber_list](https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/test_helpers/email_alert_api.rb#L27).

spec/models/email_alert_signup_spec.rb
```diff
@@ -0,0 +1,80 @@
+require "spec_helper"
+
+describe EmailAlertSignup do
+
+  let (:content_item) {
+    OpenStruct.new(
```
tekin: Would it be possible to write this spec to use the example from `govuk-content-schemas`? That way, it will always run against the latest version of an email-alert-signup, giving us coverage against breaking changes to the format.

spec/models/email_alert_signup_spec.rb
tekin: For example:

```ruby

create_subscriber_list_request = email_alert_api_creates_subscriber_list(subscriber_list_params)
expect(create_subscriber_list_request).to have_been_requested

```

spec/models/email_alert_signup_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+
+  it "is invalid with no content item" do
+    expect(EmailAlertSignup.new(nil)).not_to be_valid
+  end
+
+  describe "#save" do
+    it "creates the topic in GovDelivery using the tag and title" do
+      email_signup = EmailAlertSignup.new(content_item)
+
+      allow(email_alert_api).to receive(:find_or_create_subscriber_list).with(
```
tekin: Again, this just sets a stub, but does not follow it up with an expectation that the thing was actually called.

app/models/email_alert_signup.rb
```diff
@@ -0,0 +1,54 @@
...
+    }
+  end
+
+private
+  attr_reader :content_item
+
+  def subscription_params
+    {
+      title: title,
+      tags: tags.marshal_dump
```
tekin: Would be good to have a comment or something to indicate why `marshal_dump`. Or better yet, [extract method](http://refactoring.com/catalog/extractMethod.html)!

```
...
  tags: convert_openstruct_to_hash(tags)
...

def convert_openstruct_to_hash(open_struct)
  open_struct.marshal_dump
end

```

app/models/email_alert_signup.rb
```diff
@@ -0,0 +1,54 @@
...
+    }.deep_stringify_keys
+  end
+
+  def raw_breadcrumbs
+    if content_item.details.breadcrumbs
+      content_item.details.breadcrumbs.map(&:marshal_dump)
```
tekin:
```
content_item.details.breadcrumbs.map { |crumb| convert_openstruct_to_hash(crumb) }
```

tekin:
:tada:

PR #7, opened 2015-05-14
Use govdelivery-title if available

tommyp:
If a different govdelivery-title is present, use that for the name of
the govdelivery topic when creating the subscription params.


elliotcm:
Tests?

tommyp:
Added tests for cases of it being present.

PR #17, opened 2015-11-11
Use the email_alert_type when constructing the email-alert-api payload

mobaig:
Please merge https://github.com/alphagov/email-alert-api/pull/93 first.

This makes use of a new field on email_alert_signup content items -
'email_alert_type' - to construct the payload for the email-alert-api
when sending a subscription request. Using the field means the frontend
doesn't need to have explicit knowledge of the types of email alert that
are supported. Currently only policy subscriptions are handled in this
way, but the implementation leaves open the possibility of migrating
topics (for example) to make use of the email-alert-frontend as well.

Govuk content schemas determine which values of email_alert_type are valid,
ensuring that the email-alert-api receives expected values.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,7 +17,7 @@ gem 'decent_exposure', '~> 2.3.2'
 gem 'gds-api-adapters', '~> 20.1.1'
 
 group :development, :test do
-  gem 'byebug'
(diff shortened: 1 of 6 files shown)


rboulton:
This looks right to me.

PR #18, opened 2015-11-13
Add conditional check on subscription params tag construction

mobaig:
Email alert signup content items are currently being migrated to use the
field 'signup_tags' instead of 'tags' to describe the content item being
subscribed to. This change provides a temporary conditional check on the
presence of that field while the schema change is being reviewed and
deployed.


mobaig:
Ready to merge any time - no dependency on any other email-related app.

PR #27, opened 2017-01-06
Add Jenkinsfile

suzannehamilton:
Configure project to build on Jenkins 2.

There are a couple of environment variables from jenkins.sh which I haven't migrated, because they aren't used anywhere else in the project:

```
export GOVUK_APP_DOMAIN=test.gov.uk
export GOVUK_ASSET_ROOT=http://static.test.gov.uk
```

Does that sound OK? Or are they required by some dependency which is not in this repo?

I've also added the schema test build parameters so this can be added to the content schema project's list of downstream builds. email-alert-frontend is not currently a downstream build in Jenkins 1, but I suspect that it should be, since it relies on the [email_alert_signup](https://github.com/alphagov/govuk-content-schemas/tree/master/formats/email_alert_signup) schema.

https://trello.com/c/S4DZbsPa/285-jenkins-2-migration

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,92 @@
+#!/usr/bin/env groovy
+
+REPOSITORY = 'email-alert-frontend'
+DEFAULT_SCHEMA_BRANCH = 'deployed-to-production'
+
+node {
+  def govuk = load '/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy'
+
+  properties([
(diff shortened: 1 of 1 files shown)


cbaines:
> Does that sound OK? Or are they required by some dependency which is not in this repo?

They might be used by Plek (see https://github.com/alphagov/plek/blob/master/lib/plek.rb ), but if the tests pass, then they pass, and less configuration is probably better.

PR #30, opened 2017-02-09
Add  starting the app using the live content store

mcgoooo:
this puts the email app in line with government-front, specialist content
and a few others

same pull in other repos can be seen here
https://github.com/alphagov/specialist-frontend/pull/129
https://github.com/alphagov/contacts-frontend/pull/73


startup.sh
```diff
@@ -1,4 +1,13 @@
...
-bundle exec rails s -p 3099
+
+if [[ $1 == "--live" ]] ; then
+  GOVUK_APP_DOMAIN=www.gov.uk \
+  GOVUK_WEBSITE_ROOT=https://www.gov.uk \
+  PLEK_SERVICE_CONTENT_STORE_URI=https://www.gov.uk/api \
+  PLEK_SERVICE_STATIC_URI=assets.publishing.service.gov.uk \
+  bundle exec rails s -p 3099
+else
+  bundle exec rails s -p 3099
```
fofr: Looks like `bundle exec rails s -p 3099` can be pulled out of the if statement.

mcgoooo: Yo, this is discussed in here
https://github.com/alphagov/specialist-frontend/pull/129

PR #33, opened 2017-03-23
Taxonomy email alert signup pages

mobaig:
Add the taxonomy email alert signup pages, served from `/email-alert`.

This PR includes some initial cleaning up of the test suite and dependencies - recommend using [this diff](https://github.com/alphagov/email-alert-frontend/pull/33/files/7e9934e75b1295d535decefa61330e1de27aa064..b15d158bdef94f60f8776989ed76273cac750d37) to review the actual feature.

https://trello.com/c/lPlSbZnX/166-3-build-3-types-of-sign-up-pages-backend

## New sign up (when child taxons present)

screen shot 2017-03-28 at 11 58 28

## New sign up (when looking at leaf taxon)

screen shot 2017-03-28 at 12 03 11

## Confirm page

screen shot 2017-03-28 at 11 58 44


app/views/taxonomy_signups/new.html.erb
```diff
@@ -0,0 +1,40 @@
...
+    <fieldset>
+      <legend id="alerts" class="visually-hidden">What do you want to get alerts about?</legend>
+      <div class='email-signup-taxons' role='radiogroup' aria-labelledby='alerts'>
+        <p class="email-option">Everything published to GOV.UK about:</p>
+
+        <%= render partial: 'taxon', locals: { taxon: @taxon, is_parent: true } %>
+
+        <p class='email-option form-block'>or only one of the following</p>
+
+        <% @taxon.dig('links', 'child_taxons').each_with_index do |taxon, taxon_index| %>
```
dwhenry: might be nicer to use:

```
<%= render partial: "taxon", collection: @taxon.dig('links', 'child_taxons') %>
```
this would give youn the `index` variable in then view rather than having to pass it in.

app/controllers/taxonomy_signups_controller.rb
```diff
@@ -0,0 +1,43 @@
+class TaxonomySignupsController < ApplicationController
+  def new
+    redirect_to '/' and return unless params[:paths].present?
```
dwhenry: should we be redirecting back rather than to root?

app/views/layouts/application.html.erb
```diff
@@ -3,6 +3,7 @@
   <head>
     <title><%= yield :title %> - GOV.UK</title>
     <%= stylesheet_link_tag "application" %>
+    <%= javascript_include_tag 'application' %>
```
dwhenry: shouldn't this only be loaded in the footer? or at least lazy-loaded

app/models/taxonomy_signup.rb
```diff
@@ -0,0 +1,32 @@
...
+  def update_subscription
+    EmailAlertFrontend.services(:email_alert_api)
+      .find_or_create_subscriber_list(subscription_params)
+  end
+
+  def subscription_params
+    {
+      'title' => taxon['title'],
+      'links' => {
+        'taxons' => [taxon['content_id']]
```
mobaig: quite right, updated.

app/assets/stylesheets/application.scss
```diff
@@ -15,3 +17,5 @@
     margin: 20px 0;
   }
 }
+
+@import "taxonomy-signup"
```
mobaig: removed.

app/controllers/taxonomy_signups_controller.rb
mobaig: That would break in the absence of a referer - redirecting to root is the simplest thing to do in this case, especially as you'd never have a link on the site without the paths specified.

app/controllers/taxonomy_signups_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    taxon_path = params[:paths].first
+    load_taxon(taxon_path)
```
mobaig: Breaks noisily if it can't find a taxon with the given base path, which is reasonable i think.

app/views/layouts/application.html.erb
tijmenb: I think that doesn't work in our layouts, this is the way it's done in the frontends I've seen.

tijmenb:
Can you add some screenshots?

app/views/layouts/application.html.erb
mobaig: moved.

mobaig:
@tijmenb added

Gemfile
```diff
@@ -3,7 +3,8 @@ source 'https://rubygems.org'
 gem 'rails', '4.2.5.2'
 gem 'slimmer', '10.0.0'
 
-gem 'govuk_frontend_toolkit', '~> 3.1.0'
+gem 'govuk_elements_rails', '~> 3.0.1'
```
nickcolley: We're currently not using GOV.UK Elements on any GOV.UK Publishing frontend applications.

Seems as though the kind of UI you're implementing it makes sense to make use of Elements, just Elements is technically built in a different way to the rest of GOV.UK Publishing frontend.

Maybe let's have a chat through it offline

app/assets/javascripts/show-hide-content.js
```diff
@@ -0,0 +1,172 @@
+;(function (global) {
```
nickcolley: What's the difference between this and https://github.com/alphagov/govuk_frontend_toolkit/blob/c88df4977fbb1d63950e8d19cc7dfbd02ede0709/javascripts/govuk/show-hide-content.js ?

app/views/taxonomy_signups/new.html.erb
mobaig: not worth it in this case, less clear whats going on - particularly as you then need to handle a default value for when you're rendering the parent taxon with this partial.

nickcolley:
In your screenshots looks like there are indentation issues, the breadcrumbs should line up with the body of the page.

app/assets/javascripts/show-hide-content.js
mobaig: @nickcolley nothing, from the looks of it. The frontend work in this PR is almost entirely @carolinegreen . I would assume that with the removal of govuk_frontend_toolkit from this project, the show-hide-content javascript was simply copied across out of pragmatism.

nickcolley: @mobaig @carolinegreen did govuk_frontend_toolkit conflict with govuk_elements?

app/views/taxonomy_signups/confirm.html.erb
```diff
@@ -0,0 +1,41 @@
+<% content_for :title, @taxon['title'] %>
+
+<%= render 'govuk_component/breadcrumbs', breadcrumbs: @breadcrumbs %>
+
+<div class='email-signup column-two-thirds'>
+<div class="govuk-title length-long">
```
nickcolley: This might work with http://govuk-component-guide.herokuapp.com/components/title

app/assets/javascripts/show-hide-content.js
mobaig: @nickcolley Unfortunately @carolinegreen is on holiday this week : (.

app/controllers/taxonomy_signups_controller.rb
mobaig: @dwhenry added a controller test for this.

app/controllers/taxonomy_signups_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+    load_breadcrumbs
+  end
+
+  def confirm
+    redirect_to '/' and return unless params[:'taxon-list'].present?
+
+    taxon_path = params[:'taxon-list']
```
mobaig: Updated - I've opted to call the query param 'topic' across the controller and view. This is how we describe taxons to users so probably sensible to use that in the URL.

app/views/taxonomy_signups/confirm.html.erb
mobaig: Updated to use the component instead.

dwhenry:
As discussed would prefer not to use `topic` in the url, but understand the reasoning - nice external urls - so feel it is fine to do in this case.

app/assets/javascripts/show-hide-content.js
carolinegreen: No, I wasn't familiar with rails asset pipeline, and was making sure it worked before I put that in.  Managed to miss this, I've put in a pull request for this now.

https://github.com/alphagov/email-alert-frontend/pull/34

Gemfile
carolinegreen: As far as I know you don't have or hardly have anything like this on GOV.UK.  Looks like you had a chat about it though, so that's good.

PR #34, opened 2017-03-28
Use rails asset path

carolinegreen:
Remove duplicate javascript file.


nickcolley:
Does this include all of https://github.com/alphagov/govuk_frontend_toolkit/blob/master/javascripts/govuk_toolkit.js ? Do we need it all?

What's the impact on the application bundle?

carolinegreen:
Yes, I'm aware of that, but the current documentation seems to suggest including everything is the  preferable default despite performance costs.  I'm going to re-write some of the documentation when I'm back from holiday to try and make it clearer.  Hopefully this is now fine and can be merged.  I wouldn't usually check on things like this, but I'm aware the email team won't be around after the end of this week.

PR #35, opened 2017-03-28
Add email volume estimates to signup confirm page

mobaig:
Add very rough guesses about the weekly volume of email a user can
expect depending on which level of the taxonomy they sign up to. These
figures are based on analysis of historical email volume.

Note that the original requirement here was to treat 2nd level taxons that are also bottom level taxons as being LOW frequency - I've made a decision here to simplify the implementation because of time constraints. This is unlikely to be problematic given these numbers are very rough estimates. Discussed and agreed with Rob Rankin.

https://trello.com/c/lPlSbZnX/166-3-build-3-types-of-sign-up-pages-backend

### DIFF
--- a/app/models/weekly_email_volume.rb
+++ b/app/models/weekly_email_volume.rb
@@ -1,0 +1,35 @@
+class WeeklyEmailVolume
+  HIGH = '40-60'.freeze
+  MEDIUM = '0-20'.freeze
+  LOW = '0-5'.freeze
+
+  def initialize(taxon)
+    @taxon = taxon
+  end
+
+  def estimate
+    parent_taxon = extract_parent_from(@taxon)
+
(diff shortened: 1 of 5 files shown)


app/models/weekly_email_volume.rb
```diff
@@ -0,0 +1,36 @@
...
+  MEDIUM = '0-20'.freeze
+  LOW = '0-5'.freeze
+
+  def initialize(taxon)
+    @taxon = taxon
+  end
+
+  def estimate
+    parent_taxon = extract_parent_from(@taxon)
+    child_taxons = @taxon.dig('links', 'child_taxon')
```
dwhenry: we don't appear to use this variable anywhere

PR #43, opened 2017-04-11
Validate user input more thoroughly for taxonomy signups

jennyd:
This makes the existing checks more consistent, and adds a couple of new ones. We had some [requests over the weekend with unexpected param values](https://errbit.publishing.service.gov.uk/apps/5501a9da0da115b6b8004b79/problems/58e9474665786327ddc80b00) and it's better to not return an error response if we can avoid it.

I considered also adding a check for the [content store response being 400](https://errbit.publishing.service.gov.uk/apps/5501a9da0da115b6b8004b79/problems/58e9474765786327ed000b00) but wasn't sure if that's expected behaviour - maybe that should be fixed in content store?

### DIFF
--- a/app/controllers/taxonomy_signups_controller.rb
+++ b/app/controllers/taxonomy_signups_controller.rb
@@ -2,6 +2,7 @@ class TaxonomySignupsController < ApplicationController
   protect_from_forgery except: [:create]
   before_action :require_taxon_param
   before_action :load_taxon
+  before_action :validate_taxon_document_type
 
   def new
     load_breadcrumbs
@@ -44,6 +45,10 @@ private
     @taxon ||= EmailAlertFrontend
       .services(:content_store)
       .content_item(taxon_path)
+  end
+
+  def validate_taxon_document_type
+    redirect_to '/' and return unless @taxon['document_type'] == 'taxon'
   end
 
   def load_breadcrumbs
--- a/app/controllers/taxonomy_signups_controller.rb
+++ b/app/controllers/taxonomy_signups_controller.rb
@@ -29,7 +29,9 @@ private
   end
 
   def valid_taxon_param?
-    taxon_path.present?
+    taxon_path.to_s.starts_with?('/') && URI.parse(taxon_path).relative?
+  rescue URI::InvalidURIError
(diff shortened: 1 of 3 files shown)


spec/controllers/taxonomy_signups_controller_spec.rb
```diff
@@ -3,19 +3,65 @@
...
 
     it 'errors if no taxon found' do
       content_store_does_not_have_item('/education/some-rando-item')
-      get :new, topic: '/education/some-rando-item'
+      make_request(topic: '/education/some-rando-item')
 
       expect(response.status).to eq 404
     end
+
+    it 'redirects to root unless the content item is a taxon' do
```
dwhenry: This feels like it describes what you don't want to happen. might be nicer as ```it 'renders the response when content item is a taxon' do```

app/controllers/taxonomy_signups_controller.rb
```diff
@@ -45,6 +47,10 @@ def load_taxon
       .content_item(taxon_path)
   end
 
+  def validate_taxon_document_type
+    redirect_to '/' and return unless @taxon['document_type'] == 'taxon'
```
dwhenry: Can taxon be `nil` or in any other way be an object that does not respond to `[]` or is it just a JOSN object?

jennyd: It shouldn't be `nil`, I think we'll always have a content item by the time we get here - 404s from content store raise exceptions which somehow magically turn into a 404 response from email-alert-frontend, and any other 4xx or 5xx response from content store should also raise an exception before we get this far.

spec/controllers/taxonomy_signups_controller_spec.rb
jennyd: This test is for the case where it isn't a taxon, though... There's a [cucumber feature for the happy path](https://github.com/alphagov/email-alert-frontend/blob/2b113232be7f1c82adcb3519d38b9a37917d82da/features/taxonomy_email_alerts.feature).

dwhenry:
👍

PR #45, opened 2017-05-18
Prevent search indexes indexing email signup pages on search

theKHutDeveloper:
Added meta tag to stop email alert pages from being indexed by search engines.

Trello: https://trello.com/c/s7JviEzz/131-prevent-search-indexes-indexing-email-signup-pages-on-search

### DIFF
--- a/app/views/taxonomy_signups/new.html.erb
+++ b/app/views/taxonomy_signups/new.html.erb
@@ -1,6 +1,6 @@
 <% content_for :title, @taxon['title'] %>
 <% content_for :head do %>
-  <meta NAME="ROBOTS" CONTENT="NOINDEX, NOFOLLOW">
+  <meta name="robots" content="noindex, nofollow">
 <% end %>
 <%= render 'govuk_component/breadcrumbs', breadcrumbs: @breadcrumbs %>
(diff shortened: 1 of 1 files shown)


theKHutDeveloper:
@carvil I've added the meta tag

app/views/taxonomy_signups/new.html.erb
```diff
@@ -1,5 +1,7 @@
 <% content_for :title, @taxon['title'] %>
-
+<% content_for :head do %>
+  <meta NAME="ROBOTS" CONTENT="NOINDEX, NOFOLLOW">
```
carvil: I think this meta tag will work. Could you please change it to lowercase?

fofr:
Is the nofollow needed to prevent search engines from clicking links?

theKHutDeveloper:
@carvil Changed meta tag to lowercase

carvil:
@fofr that's correct. I don't think there are any relevant links to follow within email signup pages, so besides telling robots to not index the page (and therefore avoid double entries in search), we are also telling them to not click on anything else. Do you think that's a fair assumption?

PR #48, opened 2017-09-05
Upgrade to Rails 5.1

rubenarakelyan:
This PR upgrades email-alert-frontend to Rails 5.1.

Trello: https://trello.com/c/X9s96x2T/5-upgrade-email-alert-frontend-to-a-supported-version-of-rails


fofr:
Not sure what the changes to elements will affect. Have you done like for like comparisons to check for unexpected style differences? (ie manual visual regression tests 😕 )

On government-frontend and static we noticed that Rails 5.1 turns on asset digests in the test environment. We precompile assets on jenkins so tests continued to pass, but locally we didn't have a precompile step and we found that some JS was unavailable. Are there any JS dependencies in this app, and are they integration tested?

Otherwise it all looks fairly standard.

rubenarakelyan:
@fofr Yes, I've manually tested the pages in integration and they look identical to the production versions. There is one JS file which contains the show/hide behaviour which I have also tested manually. The JS doesn't have an explicit integration test but is implicitly tested by other feature tests since they look for a button that is only revealed by the JS.

PR #51, opened 2017-10-12
[do not merge] Consolidate finder-frontend email signups

tuzz:
To do:

- get the thing rendering
- add controller test coverage that actually renders the views
- disambiguate the routes
- name everything so it's obvious this controller serves specialist publisher only
- review / merge / deploy
- publish the signup content items at the new path
- republish the finders to point to the new path
- delete code from finder-frontend


rubenarakelyan:
Closing this for now until we get around to it again

PR #63, opened 2017-12-06
Add subscription page

gpeng:
This is a very bare bones implementation of a subscription page on email-alert-frontend that we can use to replace the GovDelivery version e.g (https://public.govdelivery.com/accounts/UKGOVUK/subscriber/new?topic_id=UKGOVUK_2332)

The journey begins.. (dun, dun, darrrrr) at `/email/subscriptions/new?topic_id=UKGOVUK_2332` and entering an email address into there and submitting the form will create a subscription for the supplied 'topic'.

Currently missing:

* the form isn't built or styled by a professional
* a 404 response (or anything other than a success) from email-alert-api when fetching the subscribable isn't handled, email-alert-frontend will 500
* handling failure to actually create the subscription

... and probably lots of other things but it does give us enough to test 'end to end' now and something to build out from.

I've restricted the routes in production so this should only be served on integration for now as that was the simplest way I could come up with.

[Trello](https://trello.com/b/luhhZk3B/govuk-email-2017-18-doing)

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -11,11 +11,9 @@ Rails.application.routes.draw do
   get '/email/unsubscribe/:uuid' => 'unsubscriptions#confirm', as: :confirm_unsubscribe
   post '/email/unsubscribe/:uuid' => 'unsubscriptions#confirmed', as: :unsubscribe
 
-  if Rails.env != "production" || ENV["GOVUK_APP_DOMAIN"] =~ /integration/
-    scope '/email' do
-      resources :subscriptions, only: %i[create new]
-      get '/subscriptions/complete' => 'subscriptions#show', as: :subscription
-    end
+  scope '/email' do
+    resources :subscriptions, only: %i[create new]
+    get '/subscriptions/complete' => 'subscriptions#show', as: :subscription
   end
 
   if Rails.env.test?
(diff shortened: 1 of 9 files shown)


config/routes.rb
```diff
@@ -11,6 +11,13 @@
   get '/email/unsubscribe/:uuid' => 'unsubscriptions#confirm', as: :confirm_unsubscribe
   post '/email/unsubscribe/:uuid' => 'unsubscriptions#confirmed', as: :unsubscribe
 
+  if Rails.env != "production" || ENV["GOVUK_APP_DOMAIN"] =~ /integration/
```
thomasleese: Do we need to have this restriction in here? Since people will only go to this URL by either typing it in or [if we enable it in Email Alert API](https://github.com/alphagov/email-alert-api/pull/284/files#diff-266065afad82bba6259d1fc97ce5612cR60).

gpeng: I thought we'd decided that we didn't want to risk people finding the page. This way they won't but we can still use it for manual end to end testing on integration. Happy to remove it if you disagree.

thomasleese: I don't really mind, I just figured it's an extra level of hiding when we've already got it at the Email Alert API level. It's not enough to stop approval though thinking about it.

tuzz:
Tested on integration. All working with the feature flag set in puppet. Merging.

PR #95, opened 2018-01-09
Style the email collection page and make it accessible

tuzz:

This PR also sets up `govuk_publishing_components`. This is the preferred way to build frontend applications to promote reuse of components across applications and improve consistency in design and accessibility.

There's still a bit more to do, but this is already a big improvement.

### Screenshots ###

screen shot 2018-01-09 at 10 00 16

screen shot 2018-01-09 at 10 00 30


spec/features/subscribe_spec.rb
```diff
@@ -30,8 +30,9 @@
       visit "/email/subscriptions/new?topic_id=#{topic_id}"
       fill_in :address, with: address
       expect(page).to have_content("Test Subscriber List")
-      click_button "Subscribe"
-      expect(page).to have_content("Subscription created successfully.")
+
+      page.execute_script("document.querySelector('form').submit()")
```
gpeng: Why do you prefer this over `click_button`?

spec/features/subscribe_spec.rb
```diff
@@ -30,8 +30,9 @@
       visit "/email/subscriptions/new?topic_id=#{topic_id}"
       fill_in :address, with: address
       expect(page).to have_content("Test Subscriber List")
-      click_button "Subscribe"
-      expect(page).to have_content("Subscription successfully created")
+
+      page.execute_script("document.querySelector('form').submit()")
```
gpeng: Why do you prefer this over `click_button`?

tuzz: I don't, but currently `govuk_publishing_components` doesn't render the real button component in a test environment, it just puts a kind of fixture in its place that looks like this:

`<test-govuk-component data-template="govuk_component-button">{"text":"Subscribe","margin_bottom":true}</test-govuk-component>`

We didn't have this problem before because we weren't using a component. It's kind of annoying, but I think this is something the #publishing-frontend team want to improve for all components eventually.

gpeng: Ok... fair enough. Can you add a comment?

PR #96, opened 2018-01-09
Add back links to collection pages as per service manual

tuzz:
https://www.gov.uk/service-manual/design/question-pages#back-link

Screenshots:
screen shot 2018-01-09 at 12 28 25
screen shot 2018-01-09 at 12 28 35

### DIFF
--- a/app/assets/stylesheets/application-without-elements.scss
+++ b/app/assets/stylesheets/application-without-elements.scss
@@ -2,6 +2,7 @@
 @import "design-patterns/buttons";
 
 // Components from govuk_publishing_components:
+@import "components/back-link";
 @import "components/label";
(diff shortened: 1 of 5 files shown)


fofr:
Can we get a review app and pipeline set up for this app on the main Heroku account?
There used to be one associated with my personal account but it's been deleted because of the changes Heroku made to pipeline ownership.

This will give us some URLs to review changes like our other frontend apps.

tuzz:
@fofr I spoke to @nickcolley about this. The problem is the app depends on `email-alert-api` and to render a page, it needs to communicate with it. As far as I know, we can't currently communicate to an internal app from Heroku.

PR #101, opened 2018-01-12
Add the error-summary component to the email collection page

tuzz:

This pattern is used elsewhere on GOV.UK. It
summarises what went wrong and links to the field
that needs to be fixed.

This element is focussed with JavaScript on page
load which means it’s the first thing screen
readers read out which is a helpful accessibility
pattern.


app/views/subscriptions/new.html.erb
```diff
@@ -10,12 +10,24 @@
...
-<%= render "components/back_link", {
+<%= render "govuk_publishing_components/components/back_link", {
   href: @back_url,
 } %>
 
 <div class='grid-row'>
   <div class='column-two-thirds'>
+    <% if flash[:error] %>
+      <%= render "govuk_publishing_components/components/error_summary", {
+        title: "This isn't a valid email address",
```
fofr: Minor: Text could use smart quotes for better readability (http://smartquotesforsmartpeople.com/)

fofr:
Looks good. Is there a screenshot of the new UI?
(#102 would make a thorough review a little easier)

app/views/subscriptions/new.html.erb
rubenarakelyan: Fixed

rubenarakelyan:
screen shot 2018-01-15 at 11 41 45 1

PR #111, opened 2018-01-29
Include frequency in steps to sign up for Email Alert API

kevindew:
https://trello.com/c/4gn2Vr6x/539-update-email-alert-frontend-to-add-frequency

This introduces a frequency step into the process for signing up for email-alert-api subscribables.

## Visual Flow

The flow to signup for an email is now like this:

**GET /email/subscriptions/new?topic_id=UKGOVUKDUP_46041**
screen shot 2018-01-29 at 12 17 56

**POST /email/subscriptions/frequency** => **GET /email/subscriptions/new?topic_id=UKGOVUKDUP_46041&frequency=daily**
screen shot 2018-01-29 at 12 29 35

**POST /email/subscriptions/create** => **GET /email/subscriptions/complete?topic_id=UKGOVUKDUP_46041&frequency=daily**
screen shot 2018-01-29 at 12 29 41

## Supplementary changes

- Redirect when no referer now goes to website root from Plek rather than hardcoding production gov.uk
- Now has different errors for missing and invalid email address
- No longer hardcodes invalid email error in alert box of page
- Changed the route for when task is complete to be /email/subscriptions/complete rather than /email/subscriptions
- Using curly quotes in text rendered to user

## Follow up tasks

https://trello.com/c/vkFPJftw/380-improve-communications-and-visuals-for-email-alert-sign-up


alanmaddrell:
About reducing the number of pages in this journey, the surprising insight from testing of different services is that users prefer having one simple thing per page, even if it means more pages. Seems almost counterintuitive but it does make things simpler. +1 to Bevan's comment about the radio buttons

kevindew:
@gpeng 
> Having the frequency and email bits on separate views seems to add a lot of complexity in extra actions, validation etc

Yeah it would be pretty trivial for this to be on one page, and it all feels a bit clunky doing a multi-step form without a session. Though I went this way as it's consistent with the prototype, and It's all referenced in the [service manual](https://www.gov.uk/service-manual/design/form-structure) (as per the comment from @alanmaddrell)

@bevanloon @alanmaddrell 
> In the first screenshot , the radio buttons and the "Next" button look quite squashed together - is it possible to add a little space between them - for example https://www.tax.service.gov.uk/government-gateway-registration-frontend/are-you-trying-to-file-for-sa?continue=%2Fcheck-your-state-pension%2Faccount&origin=nisp-frontend has a bit more space between tte elements

Yeah that is meant to be something to resolve in the follow up card - apologies if that isn't clear enough - I figured getting the content into shape might raise more styling issues and that it'd be better to get resolved in one go.

I'm ignorant of how we tackle CSS for GOV.UK so although it could be made consistent with the prototype rather trivially by adding a `margin-top: 35px` this didn't really feel the way to do it. I tried using the class from your example @bevan on fieldset of form-field-group which pads that out but it looks like it's one they created. So I'm kinda wanting to get insight from someone more adept GOV.UK CSS to either show what was missed so this snaps into place, or to say bung an extra class in.

bevanloon:
@kevindew  👍 perhaps we should rethink how we've implemented our own components for input (https://github.com/alphagov/email-alert-frontend/blob/master/app/views/components/_input.html.erb) and confirmation (https://github.com/alphagov/email-alert-frontend/blob/master/app/views/components/_confirmation_box.html.erb) if these exist in `govuk_publishing_components` otherwise we have inconsistencies in the way we are doing the frontend work in this project.

kevindew:
@bevanloon sounds good. I'll add that to the trello card. Are you happy for this to be merged?

PR #139, opened 2018-02-21
Use the feedback component

tijmenb:
This replaces the "Is there anything wrong?" link with the new feedback component. This component asks the user if the page is useful and reports the results to Google Analytics.

See https://github.com/alphagov/govuk_publishing_components/pull/163 for the component.

https://trello.com/c/eKtXGtBD

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,7 @@ gem 'rails', '5.1.5'
 gem 'decent_exposure', '~> 3.0'
 gem 'foreman', '~> 0.84.0'
 gem 'sass-rails', '~> 5.0'
-gem 'slimmer', '~> 11.0'
+gem 'slimmer', '~> 12.0'
 gem 'uglifier', '~> 4.1'
 
 gem 'gds-api-adapters', '~> 51.1'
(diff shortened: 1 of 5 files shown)


kevindew:
This all looks good to me. I do notice though that we quite a small inconsistency between a couple of our pages.

We have a bunch of pages that use a different stylesheet [application-without-elements.scss](https://github.com/alphagov/email-alert-frontend/blob/master/app/assets/stylesheets/application-without-elements.scss) than application.scss and on those this renders with a smaller font (due to [elements typography](https://github.com/alphagov/govuk_elements/blob/master/assets/sass/elements/_elements-typography.scss#L7-L10)) 

So on a page of url: /email/subscriptions/new?topic_id=UKGOVUK_1 this component is registered like this:
screen shot 2018-02-21 at 19 06 07

as opposed to:
screen shot 2018-02-21 at 19 07 28

I couldn't find info in the git log as to why we've got two diverging stylesheets unfortunately.

I imagine if we could apply a rule like govuk_elements this could go away (and probably fix a couple of other consistency issues we've got ourselves) but flagging this all up since I know very little about our frontend and figure you may hit this in other projects.

rubenarakelyan:
Yeah, we need to move email-alert-frontend to use GOV.UK components properly and without overriding stylesheets - there is a piece of work to consolidate the various email frontends that would fit in nicely with this

tijmenb:
We'll look into making the component more resistant against collisions and bump the version once that's released.

I've captured the elements thing on the tech debt board: https://trello.com/c/klZnFNZi

PR #177, opened 2018-03-21
Set Cache-Control header and allow CSRF protection

rubenarakelyan:
This commit sets the `Cache-Control` header for all controllers that serve pages under `/email` to “private” so they don’t get cached by the CDN. This is because cookies and sessions will be enabled soon and we don’t want specific users’ pages to be cached. It also enables CSRF protection now that cookies are enabled.

Depends on https://github.com/alphagov/govuk-puppet/pull/7402

Trello: https://trello.com/c/97eS4TdI/710-sort-out-sessions-for-email-alert-frontend

### DIFF
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -1,6 +1,5 @@
 class SubscriptionsController < ApplicationController
-  protect_from_forgery except: %i[create frequency]
-
+  before_action :set_cache_control_header
   before_action :assign_attributes
   before_action :assign_back_url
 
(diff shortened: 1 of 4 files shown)


kevindew:
Do you know if we need these cache control headers? It looks like we already send private header by default:

```
email-alert-api git:(email-status-indexes) curl -I -s https://www.gov.uk/email/subscriptions/new\?topic_id\=all-publications-about-all-topics-by-all-organisations | grep Cache-Control
Cache-Control: max-age=0, private, must-revalidate
```

rubenarakelyan:
I think that's the rack default (https://github.com/rack/rack/blob/master/lib/rack/etag.rb#L16) but I thought it would be prudent to set it explicitly in case some rack or nginx config changes and we no longer set it by default.

PR #183, opened 2018-03-22
Add subscription management frontend

rubenarakelyan:
This PR adds the subscription management frontend, which allows subscribers to view their subscriptions, change their email address, change the frequency of a subscription, unsubscribe from a list or unsubscribe from everything.

Trello: https://trello.com/c/zy8FYa0N/722-add-a-page-to-show-all-subscriptions-for-a-subscriber

Sign in:
sign-in

Link sent:
link-sent

Link expired error:
link-expired

Home:
homepage

Change email address:
change-email-address

Change email address (with error message):
change-email-address-error

Change frequency:
change-frequency

Unsubscribe from a topic:
unsubscribe

Successfully unsubscribed from a topic:
unsubscribed

Unsubscribe from everything:
unsubscribe-all

Successfully unsubscribed from everything:
unsubscribed-all

Home with no subscriptions:
homepage-not-subscribed

### DIFF
--- a/app/controllers/authentication_controller.rb
+++ b/app/controllers/authentication_controller.rb
@@ -1,0 +1,83 @@
+class AuthenticationController < ApplicationController
+  before_action :check_if_enabled
+
+  MISSING_EMAIL_ERROR = 'Please enter your email address.'.freeze
+  INVALID_EMAIL_ERROR = 'This doesn’t look like a valid email address – check you’ve entered it correctly.'.freeze
+
+  def sign_in
+    @address = params[:address]
+  end
+
+  def request_sign_in_token
+    unless params[:address].present?
+      flash.now[:error] = MISSING_EMAIL_ERROR
+      return render :sign_in
+    end
+
+    @address = params.require(:address)
+
+    email_alert_api.create_auth_token(
+      address: @address,
+      destination: process_sign_in_token_path
+    )
+  rescue GdsApi::HTTPUnprocessableEntity
+    flash.now[:error] = INVALID_EMAIL_ERROR
+    render :sign_in
+  rescue GdsApi::HTTPNotFound
+    # User isn't subscribed, but we carry on as if they were so we
+    # don't reveal this information.
+    nil
+  end
+
+  def process_sign_in_token
+    subscriber_id, redirect = read_token(params.require(:token))
+
+    if subscriber_id.nil?
+      @sign_in_link = sign_in_path
+      return render :expired_link
+    end
+
+    authenticate_subscriber(subscriber_id)
+
+    destination = safe_redirect_destination(redirect) || list_subscriptions_path
+    redirect_to destination
+  end
+
+private
+
+  def check_if_enabled
+    render plain: '404 not found', status: :not_found unless ENV['SUBSCRIPTION_MANAGEMENT_ENABLED']
+  end
+
+  def read_token(token)
+    payload, = JWT.decode(token, secret, true, algorithm: 'HS256')
+    data = payload.fetch('data')
+    subscriber_id = data.fetch('subscriber_id')
+    redirect = data.fetch('redirect')
+    [subscriber_id, redirect]
+  rescue JWT::ExpiredSignature, KeyError
+    return []
+  end
+
+  def authenticate_subscriber(subscriber_id)
+    session['authentication'] = {
+      'subscriber_id' => subscriber_id
+    }
+  end
+
+  def safe_redirect_destination(redirect)
+    return nil unless redirect
+    parsed = URI.parse(redirect)
+    redirect if parsed.relative? && redirect[0] == '/'
+  rescue URI::InvalidURIError
+    nil
+  end
+
+  def secret
+    Rails.application.secrets.email_alert_auth_token
+  end
+
+  def email_alert_api
+    EmailAlertFrontend.services(:email_alert_api_with_no_caching)
+  end
+end
--- a/app/controllers/subscriptions_management_controller.rb
+++ b/app/controllers/subscriptions_management_controller.rb
@@ -1,0 +1,96 @@
+class SubscriptionsManagementController < ApplicationController
+  before_action :check_if_enabled
+  before_action :require_authentication
+  before_action :get_subscription_details
+  before_action :set_back_url
+
+  MISSING_EMAIL_ERROR = 'Please enter your email address.'.freeze
+  INVALID_EMAIL_ERROR = 'That email address isn’t valid, or it’s already in use – check you’ve typed in your email address correctly.'.freeze
+
+  def index; end
+
+  def update_frequency
+    @subscription_id = params.require(:id)
+
+    return render status: :not_found, plain: '404 not found' if @subscriptions[@subscription_id].nil?
+
+    @current_frequency = @subscriptions[@subscription_id]['frequency']
+  end
+
+  def change_frequency
+    email_alert_api.change_subscription(
+      id: params.require(:id),
+      frequency: params.require(:new_frequency)
+    )
+
+    redirect_to list_subscriptions_path
+  end
+
+  def update_address
+    @address = @subscriber['address']
+  end
+
+  def change_address
+    @address = @subscriber['address']
+
+    unless params[:new_address].present?
+      flash.now[:error] = MISSING_EMAIL_ERROR
+      return render :update_address
+    end
+
+    new_address = params.require(:new_address)
+
+    email_alert_api.change_subscriber(
+      id: subscriber_id,
+      new_address: new_address
+    )
+
+    redirect_to list_subscriptions_path
+  rescue GdsApi::HTTPUnprocessableEntity
+    @new_address = new_address
+    flash.now[:error] = INVALID_EMAIL_ERROR
+    render :update_address
+  end
+
+  def confirm_unsubscribe_all; end
+
+  def confirmed_unsubscribe_all
+    begin
+      email_alert_api.unsubscribe_subscriber(subscriber_id)
+    rescue GdsApi::HTTPNotFound
+      # The user has already unsubscribed.
+      nil
+    end
+  end
+
+private
+
+  def check_if_enabled
+    render plain: '404 not found', status: :not_found unless ENV['SUBSCRIPTION_MANAGEMENT_ENABLED']
+  end
+
+  def subscriber_id
+    session['authentication']['subscriber_id']
+  end
+
+  def get_subscription_details
+    subscription_details = email_alert_api.get_subscriptions(
+      id: subscriber_id
+    )
+
+    @subscriber = subscription_details['subscriber']
+    @subscriptions = {}
+
+    subscription_details['subscriptions'].each do |subscription|
+      @subscriptions[subscription['id']] = subscription
+    end
+  end
+
+  def set_back_url
+    @back_url = list_subscriptions_path
+  end
+
+  def email_alert_api
+    EmailAlertFrontend.services(:email_alert_api_with_no_caching)
+  end
+end
--- a/app/views/components/_heading.html.erb
+++ b/app/views/components/_heading.html.erb
@@ -1,0 +1,11 @@
+<%
+  id = "id=#{id}" if id
+  heading_level ||= 1
+  heading_level_tag = "h#{heading_level}" if [1,2,3,4,5,6].include? heading_level
+%>
+<<%= heading_level_tag %>
+  <%= id %>
+  class="app-c-heading"
+>
+  <%= text %>
+</<%= heading_level_tag %>>
--- a/app/views/components/docs/heading.yml
+++ b/app/views/components/docs/heading.yml
@@ -1,0 +1,32 @@
+name: Heading
+description: A text heading
+body: |
+  A heading tag with an optional id attribute, used predominantly to the left of content on consultations and publications.
+
+  Real world examples:
+
+  - [Publication](https://www.gov.uk/government/publications/recognising-the-terrorist-threat)
+  - [Consultation](https://www.gov.uk/government/consultations/proposal-for-the-future-of-rotherham-goldthorpe-jobcentre)
+accessibility_criteria: |
+  The heading must:
+
+  - be part of a correct heading structure for a page
+  - be semantically represented as a heading
+  - convey the heading level
+examples:
+  default:
+    data:
+      text: 'Download the full outcome'
+  right_to_left:
+    data:
+      text: 'مستندات'
+    context:
+      right_to_left: true
+  with_id_attribute:
+    data:
+      text: 'Detail of outcome'
+      id: 'detail_of_outcome'
+  specific_heading_level:
+    data:
+      text: 'Original consultation'
+      heading_level: 3
--- a/app/views/subscriptions_management/confirm_unsubscribe_all.html.erb
+++ b/app/views/subscriptions_management/confirm_unsubscribe_all.html.erb
@@ -30,7 +30,7 @@
         margin_bottom: true,
         data_attributes: {
           module: 'auto-track-event',
-          track_category: 'email',
+          track_category: 'email_subscriptions',
           track_action: 'unsubscribe_all',
         },
       } %>
--- a/app/views/subscriptions_management/confirm_unsubscribe_all.html.erb
+++ b/app/views/subscriptions_management/confirm_unsubscribe_all.html.erb
@@ -26,8 +26,13 @@
     <%= form_tag(action: :confirmed_unsubscribe_all) do %>
       <%= hidden_field_tag(:from, @from) %>
       <%= render 'govuk_component/button', {
-       text: 'Unsubscribe',
-       margin_bottom: true,
+        text: 'Unsubscribe',
+        margin_bottom: true,
+        data_attributes: {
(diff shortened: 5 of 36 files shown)


app/controllers/subscriptions_management_controller.rb
```diff
@@ -0,0 +1,150 @@
...
+    @current_frequency = @subscriptions[@subscription_id]['frequency']
+  end
+
+  def change_frequency
+    # TODO: Enable this
+    # email_alert_api.change_subscription(
+    #   id: subscription_id,
+    #   frequency: new_frequency
+    # )
+    session[:subscriptions][subscription_id]['frequency'] = new_frequency
```
bevanloon: I realise this is stubbed out, but just to note that when we change frequency we should probably create a new subscription row in the database and set the current subscription's `ended_at` and `ended_at_reason`.

app/controllers/subscriptions_management_controller.rb
```diff
@@ -0,0 +1,149 @@
...
+
+    redirect_to list_subscriptions_path
+  end
+
+  def update_address
+    @back_url = list_subscriptions_path
+    @address = @subscriber['address']
+  end
+
+  def change_address
```
bevanloon: When we update an email address, is the current plan to just update with no confirmation or are we going to set up a confirmation link to both old and new email addresses before we action it? If we go the paranoid route then changing an address should probably invalidate the session and tell them to do the confirmation dance....

app/views/subscriptions_management/update_address.html.erb
```diff
@@ -0,0 +1,59 @@
...
+        label: { text: 'What’s your new email address?' },
+        name: :new_address,
+        type: 'email',
+        value: @new_address,
+      } %>
+      <%= render 'govuk_component/button', {
+        text: 'Save',
+        margin_bottom: true,
+      } %>
+      or <a href="<%= list_subscriptions_path %>">cancel</a>
```
bevanloon: can we use `@back_url` here instead? and also on https://github.com/alphagov/email-alert-frontend/pull/183/commits/35273141884a425cf1e71ff7b69462b99b9c5c16#diff-86da3a3fe96392fa8e55940bc662f3a3R56

app/controllers/subscriptions_management_controller.rb
```diff
@@ -0,0 +1,149 @@
...
+  before_action :check_if_access_is_enabled
+  before_action :get_subscriptions
+
+  MISSING_EMAIL_ERROR = "Please enter your email address.".freeze
+  INVALID_EMAIL_ERROR = "This doesn’t look like a valid email address – check you’ve entered it correctly.".freeze
+
+  def index; end
+
+  def update_frequency
+    @back_url = list_subscriptions_path
```
bevanloon: Is it worth setting this in a `before_action` or are we anticipating different `back_urls` depending on the action?

app/controllers/subscriptions_management_controller.rb
```diff
@@ -0,0 +1,149 @@
...
+
+  def set_cache_control_header
+    headers["Cache-Control"] = "private"
+  end
+
+  def get_subscriptions
+    # TODO: Hardcoded for now
+    #email_alert_api.get_subscriptions(id: subscriber_id)
+    if session[:subscriber].nil? || session[:subscriptions].nil?
+      @_subscription_details = {
```
kevindew: Any reason why this is set as a instance variable? - looks like it's only used in this method.

app/controllers/subscriptions_management_controller.rb
```diff
@@ -0,0 +1,149 @@
...
+        "subscriber": {
+          "address": "hong@jabra.com"
```
kevindew: 😆

rubenarakelyan: Haha yeah, unfortunately removed now for the real thing

app/controllers/subscriptions_management_controller.rb
rubenarakelyan: Leftover from a previous incarnation - fixed

app/controllers/subscriptions_management_controller.rb
rubenarakelyan: They're all the same so I've changed it to a `before_action`

app/views/subscriptions_management/update_address.html.erb
rubenarakelyan: Done

app/views/components/docs/heading.yml
```diff
@@ -0,0 +1,32 @@
+name: Heading
+description: A text heading
+body: |
+  A heading tag with an optional id attribute, used predominantly to the left of content on consultations and publications.
+
+  Real world examples:
+
+  - [Publication](/government/publications/recognising-the-terrorist-threat)
+  - [Consultation](/government/consultations/proposal-for-the-future-of-rotherham-goldthorpe-jobcentre)
```
kevindew: These links 404 - I don't think component guide is normally (or ever) on GOV.UK so would need fully qualified links

rubenarakelyan: Fixed

app/controllers/subscriptions_management_controller.rb
rubenarakelyan: I was going down the route of just changing it...

app/views/components/_heading.html.erb
```diff
@@ -0,0 +1,11 @@
+<%
```
fofr: Is this heading component identical to https://government-frontend.herokuapp.com/component-guide/heading?
It looks very similar. If it is, is there a plan to move the component into the publishing components so the code is shared?

app/views/subscriptions_management/confirm_unsubscribe_all.html.erb
```diff
@@ -0,0 +1,36 @@
...
+  <div class="column-two-thirds">
+    <%= render 'govuk_component/title', title: 'Are you sure you want to unsubscribe from everything?' %>
+  </div>
+</div>
+
+<div class="grid-row">
+  <div class="column-two-thirds">
+    <%= render 'govuk_component/govspeak', {
+      content: %{<p>You won’t get any more automated emails from GOV.UK.</p>},
+      rich_govspeak: true
```
fofr: `rich_govspeak` only enables bold, doesn't look like there's any bold in the content.

app/views/components/_heading.html.erb
rubenarakelyan: It is indeed the same thing that I copied over. I was going to move it into the publishing components but then I also wondered whether all this was required for a heading tag!

app/views/subscriptions_management/confirm_unsubscribe_all.html.erb
rubenarakelyan: Ah I see! I'll change it to normal govspeak

app/views/subscriptions_management/update_address.html.erb
```diff
@@ -0,0 +1,59 @@
...
+
+<div class="grid-row">
+  <div class="column-two-thirds">
+    <%= render 'govuk_component/govspeak', {
+      content: %{<p>Your current email address is #{@address}</p>},
+      rich_govspeak: true
+    } %>
+    <%= form_tag change_address_path, method: :post do %>
+      <%= render 'govuk_publishing_components/components/input', {
+        error_message: flash[:error],
```
fofr: Aside: I'd expect this error to render in red rather than grey (based on the screenshot). https://govuk-publishing-components.herokuapp.com/component-guide/input shows grey, so probably a component issue.

app/views/subscriptions_management/update_address.html.erb
```diff
@@ -0,0 +1,58 @@
...
+        label: { text: 'What’s your new email address?' },
+        name: :new_address,
+        type: 'email',
+        value: @new_address,
+      } %>
+      <%= render 'govuk_component/button', {
+        text: 'Save',
+        margin_bottom: true,
+      } %>
+      or <a href="<%= @back_url %>">cancel</a>
```
fofr: The cancel link should be below the save button. This is for accessibility reasons: screen magnifier users are less likely to see the cancel option alongside the button. This positioning should also reduce the likelihood of clicking one in error.

rubenarakelyan: I've moved the link to be in the `info_text` of the button since it's related - let me know if this isn't the right thing to do

fofr: How does that look?

app/controllers/subscriptions_management_controller.rb
```diff
@@ -0,0 +1,125 @@
...
+private
+
+  def check_if_access_is_enabled
+    # TODO: Remove this once we're ready to go live
+    render plain: '404 Not Found', status: 404 unless ENV['SUBSCRIPTION_MANAGEMENT_ENABLED']
+  end
+
+  def set_cache_control_header
+    headers['Cache-Control'] = 'private'
+  end
```
kevindew: If we're going to be using this a lot I feel like it should be in the application controller - could even be a default action that you skip if you don't want it.

app/controllers/subscriptions_management_controller.rb
```diff
@@ -0,0 +1,125 @@
...
+    subscription_params.require(:id)
+  end
+
+  def frequency
+    subscription_params.require(:frequency)
+  end
+
+  def new_frequency
+    subscription_params.require(:new_frequency)
+  end
```
kevindew: For these methods you might find it's actually more readable to just call them directly in the code rather than having convenience methods. Can get a bit confusing with a controller responsible for multiple actions.

kevindew:
FYI: I've got https://github.com/alphagov/email-alert-frontend/tree/sign-in-merge which is a merge of the sign-in-spike and this, all done kinda quick just so it can be tested on integration. It may well save you some work on this though.

rubenarakelyan:
@kevindew I've already merged your sign in work into this PR, I'll deploy it to integration this morning

app/controllers/authentication_controller.rb
```diff
@@ -0,0 +1,81 @@
...
+      redirect: list_subscriptions_path
+    )
+  end
+
+  def process_token
+    data = read_token(params.require(:token))
+
+    if data.empty?
+      @sign_in_link = sign_in_path
+      render :expired_link
```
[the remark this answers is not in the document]
rubenarakelyan: We don't really have a component for showing nice flash messages hence the separate view. I've changed it to `return render :expired_link` though.

app/controllers/subscriptions_management_controller.rb
```diff
@@ -0,0 +1,125 @@
...
+      render :update_address
+      return
+    end
+
+    begin
+      email_alert_api.change_subscriber(
+        id: subscriber_id,
+        new_address: new_address
+      )
+    rescue GdsApi::HTTPUnprocessableEntity
```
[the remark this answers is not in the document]
rubenarakelyan: Done

config/routes.rb
```diff
@@ -12,11 +12,28 @@
     get '/unsubscribe/:uuid' => 'unsubscriptions#confirm', as: :confirm_unsubscribe
     post '/unsubscribe/:uuid' => 'unsubscriptions#confirmed', as: :unsubscribe
 
+    scope '/manage' do
+      get '/' => 'subscriptions_management#index', as: :list_subscriptions
+      get '/frequency/:id' => 'subscriptions_management#update_frequency', as: :update_frequency
+      post '/change-frequency' => 'subscriptions_management#change_frequency', as: :change_frequency
```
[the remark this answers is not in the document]
rubenarakelyan: Done and also with `/address/change`

config/routes.rb
```diff
@@ -12,11 +12,28 @@
...
     scope '/subscriptions' do
       get '/new' => 'subscriptions#new', as: :new_subscription
       post '/frequency' => 'subscriptions#frequency', as: :subscription_frequency
       post '/create' => 'subscriptions#create', as: :create_subscription
       get '/complete' => 'subscriptions#complete', as: :subscription
     end
+
+    get '/authenticate' => 'authentication#sign_in', as: :sign_in
+    post '/authenticate' => 'authentication#request_token', as: :request_sign_in_token
+    get '/authenticate/:token' => 'authentication#process_token',
```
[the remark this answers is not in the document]
rubenarakelyan: OK I've changed that now

app/controllers/subscriptions_management_controller.rb
rubenarakelyan: I've moved this into the application controller since it makes sense to me that none of the subscribe/unsubscribe/manage journey should really be cached.

app/controllers/subscriptions_management_controller.rb
rubenarakelyan: Done

app/controllers/authentication_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+  end
+
+private
+
+  def check_if_enabled
+    return render plain: '404 not found', status: :not_found unless ENV['SUBSCRIPTION_MANAGEMENT_ENABLED']
+  end
+
+  def read_token(token)
+    payload, = JWT.decode(token, secret, true, algorithm: 'HS256')
```
1pretz1: There's a comma after payload ?

rubenarakelyan: Yup, that's a non-intuitive way of saying that `JWT.decode` returns an array of values and we're only interested in the first one.

1pretz1: Ahh okay!

kevindew: yeah probably easier to read if we do `payload, _ = JWT.decode` - do you find that easier @1pretz1 ? 

I think I wrote this originally, can be hard to remember what syntax is a bit deceptive.

rubenarakelyan: Yeah it was originally like that but the linter complained about it

kevindew: oh - bad linter!

app/controllers/authentication_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+    subscriber_id, redirect = read_token(params.require(:token))
+
+    if subscriber_id.nil?
+      @sign_in_link = sign_in_path
+      return render :expired_link
+    end
+
+    authenticate_subscriber(subscriber_id)
+
+    destination = safe_redirect_destination(redirect) || '/email/manage'
```
kevindew: We should change this to use a rails route helper method I think it's `list_subscriptions_path`

app/controllers/subscriptions_management_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+    email_alert_api.change_subscriber(
+      id: subscriber_id,
+      new_address: new_address
+    )
+
+    redirect_to list_subscriptions_path
+  rescue GdsApi::HTTPUnprocessableEntity
+    @new_address = new_address
+    flash.now[:error] = INVALID_EMAIL_ERROR
+    return render :update_address
```
kevindew: nitpick - doesn't need a return here.

app/controllers/authentication_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+    redirect_to destination
+  end
+
+private
+
+  def check_if_enabled
+    return render plain: '404 not found', status: :not_found unless ENV['SUBSCRIPTION_MANAGEMENT_ENABLED']
```
kevindew: Nitpick return probably not needed

app/controllers/authentication_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+    email_alert_api.create_auth_token(
+      address: @address,
+      destination: process_sign_in_token_path
+    )
+  rescue GdsApi::HTTPUnprocessableEntity
+    flash.now[:error] = INVALID_EMAIL_ERROR
+    return render :sign_in
+  rescue GdsApi::HTTPNotFound
+    # User isn't subscribed, but we carry on as if they were so we
+    # don't reveal this information.
```
kevindew: 👍

app/controllers/authentication_controller.rb
rubenarakelyan: Done

app/controllers/subscriptions_management_controller.rb
rubenarakelyan: Fixed

app/controllers/authentication_controller.rb
rubenarakelyan: Fixed

spec/controllers/subscriptions_management_controller_spec.rb
```diff
@@ -0,0 +1,282 @@
...
+  let(:subscription_id) { SecureRandom.uuid }
+  let(:new_frequency) { "weekly" }
+  let(:new_address) { "test2@example.com" }
+  let(:session_data) do
+    {
+      "authentication": {
+        "subscriber_id": subscriber_id,
+        "redirect": "/email/manage"
+      }
+    }.with_indifferent_access
```
kevindew: ooh didn't know you could call this on hashes.

PR #190, opened 2018-03-28
Show expired token error on sign-in form

kevindew:
Trello: https://trello.com/c/QJ4DK2YQ/740-add-flashes-to-subscription-management

This removes the expired link view and instead shows the contents of
that on the sign in form. It does this by using the error summary
component.

This also deactivates any subscriber when the process token action is
invoked.

Image of this in action: 
screen shot 2018-03-28 at 18 15 44


rubenarakelyan:
We should keep the same wording as before for consistency, but other than that LGTM

PR #191, opened 2018-03-28
Add flash messages

kevindew:
This adds flash messages to various operations as part of subscription management. It also adds a page to show when a subscription has ended

### Change Frequency

screen shot 2018-03-29 at 16 08 18

### Change Email

screen shot 2018-03-29 at 16 07 03

### Unsubscribe All

screen shot 2018-03-29 at 14 35 52

### Unsubscribe single

screen shot 2018-03-29 at 14 35 03

### Subscription ended page

screen shot 2018-03-29 at 16 08 44

This is a DNM as it would need https://github.com/alphagov/govuk_publishing_components/pull/254 merging


kevindew:
@fofr Looking at them today I think they probably should be moved up above the "subscriptions for test@example.com" and below "Manage your subscriptions"

app/controllers/subscriptions_management_controller.rb
```diff
@@ -17,11 +17,16 @@ def update_frequency
...
   def change_frequency
+    id = params.require(:id)
+
     email_alert_api.change_subscription(
-      id: params.require(:id),
+      id: id,
       frequency: params.require(:new_frequency)
     )
 
+    subscription_title = @subscriptions[id]['subscriber_list']['title']
```
thomasleese: Double quotes :)

app/controllers/unsubscriptions_controller.rb
```diff
@@ -1,40 +1,49 @@
...
+    @id = params.require(:id)
+    @subscription = api.get_subscription(@id)
+    @title = @subscription.dig("subscription", "subscriber_list", "title").presence
+    @authenticated_for_subscription = check_authenticated(@subscription)
   end
 
-  def from
-    params.permit(:from).fetch(:from, "")
+  def subscription_ended?
+    @subscription.dig("subscription", "ended_at").presence != nil
```
thomasleese: I think you can just do `.present?` here.

kevindew:
Thanks @thomasleese think it's mostly a single quote file there so I fixed some of the double quotes.

PR #202, opened 2018-04-11
[Don't merge] Don't use internal_name of taxons

tijmenb:
The internal_name is not meant for public consumption, it's used in content tagger to give more context to the taxonomists.

This makes it so the subscriptions are created with just the title.

The internal_name was introduced in https://github.com/alphagov/email-alert-frontend/commit/78fdf94e64f61c1f86bb91fc4aadfc5ff602aa65.

The reasoning behind it is that the titles needed to be unique in the system, something that was enforced by GovDelivery. Because we're no longer using GovDelivery, we should be able to remove the uniqueness constraint. 

A PR on email-alert-api is forthcoming, don't merge until that is deployed.

### DIFF
--- a/app/models/taxonomy_signup.rb
+++ b/app/models/taxonomy_signup.rb
@@ -22,16 +22,12 @@ private
 
   def subscription_params
     {
-      'title' => title,
+      'title' => taxon['title'],
       'links' => {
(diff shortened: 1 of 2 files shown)


tijmenb:
Closing for a bit until the email-alert-api thing is ready.

PR #219, opened 2018-05-03
Use button component from gem

andysellick:
Button component recently moved from static to the components gem. This PR updates the use of the component in this app.

The version of the gem containing the button was already incremented in a previous PR.


sihugh:
I'll deploy this now, and then I guess we should look at kibana before removing anything from static

PR #236, opened 2018-06-08
Use title component from the gem

tijmenb:
The title component is moving from Static to the gem. This makes all pages use the gem-component using a helper, which cleans up the views a bit.

https://trello.com/c/EruCrkqO

### DIFF
--- a/app/views/taxonomy_signups/confirm.html.erb
+++ b/app/views/taxonomy_signups/confirm.html.erb
@@ -7,10 +7,7 @@
 <div class='grid-row'>
 <div class='email-signup column-two-thirds'>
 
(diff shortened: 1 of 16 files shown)


app/views/taxonomy_signups/confirm.html.erb
```diff
@@ -7,10 +7,7 @@
 <div class='grid-row'>
 <div class='email-signup column-two-thirds'>
 
-  <%= render partial: 'govuk_component/title', locals: {
-    average_title_length: "long",
-    title: "What you'll get"
-  } %>
+  <%= title "What you'll get", average_title_length: "long" %>
```
rubenarakelyan: Good opportunity to change "'" to "’" :)

PR #242, opened 2018-06-18
Revert "Support subscriptions to associated taxons"

cbaines:
Internal names are by definition internal, so we shouldn't be using
these as user visible titles for subscriptions.

The original motivation for this related to GovDelivery, which is no
longer in use.

As far as I understand, there looks to be a general issue with email
subscription lists in the email-alert-api service having a unique
constraint on the title, which will probably cause some issues with
Topic Taxonomy subscriptions as the titles of taxons are not unique.

This reverts commit 78fdf94e64f61c1f86bb91fc4aadfc5ff602aa65.

### DIFF
--- a/app/models/taxonomy_signup.rb
+++ b/app/models/taxonomy_signup.rb
@@ -22,16 +22,12 @@ private
 
   def subscription_params
     {
-      'title' => title,
+      'title' => taxon['title'],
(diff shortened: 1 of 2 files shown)


cbaines:
Thanks for taking a look Kevin :+1:

PR #383, opened 2018-12-28
Refactor and clean up frontend code

alex-ju:
This is a code refactoring PR aiming to fix visual styles and make this project easier to maintain:
 - fixes the problems mentioned in [this tech debt card](https://trello.com/c/klZnFNZi) (including #39)
 - removes dependencies on `govuk_elements_rails` and `govuk_frontend_toolkit`
 - uses govuk_publishing_components and removes unnecessary components and styles
 - updates page layouts

<details>
<summary>Screen captures</summary>

taxonomy_signups/index

00_taxonomy_signups_index

taxonomy_signups/index (multiple)

01_taxonomy_signups_multiple

taxonomy_signups/confirm

02_taxonomy_signups_confirm

subscriptions/new_frequency

03_subscriptions_frequency

subscriptions/new_address

04_subscriptions_address

subscriptions/complete

05_subscriptions_success

authentication/sign_in

06_authentication

authentication/request_sign_in_token

07_authentication_request_token

</details>

---
[Trello card](https://trello.com/c/Lqs4a0p4)

PR #384, opened 2019-01-02
Fix frontend after clean up

alex-ju:
This PR fixes some bits after the major layout update in #383 as follows:
- Update email sign-up title spacing and add missing paragraph class (see captures below) - a page I didn't manage to test in local dev env
- Removes redundant output for the "Unsubscribe from everything" link
- Renders back link consistently across pages (the link was rendered in one view instead of being passed to the main layout)

### Before
old-email-subscription

### After
new-email-subscription

---
[Trello card](https://trello.com/c/Lqs4a0p4)

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -6,3 +6,12 @@
   margin-bottom: $govuk-gutter-half;
   line-height: 1.5;
 }
+
+// TODO: remove these overwrites after the title component is updated to work
+// nicely with govuk-frontend layout
+.app-title {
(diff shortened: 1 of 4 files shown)


alex-ju:
@cbaines I did an exhaustive manual check coming from different gov.uk pages and fund 3 things that needed a fix. Since you helped with the first one, would you mind reviewing this update too?

cbaines:
> @cbaines I did an exhaustive manual check on production and fund 3 things that needed a fix. Since you helped with the first one, would you mind reviewing this update too?

No problem :+1:

PR #437, opened 2019-03-25
Don't show links to alpha taxons

sihugh:
Alpha taxons are currently being used to provide single page email subscriptions
to support Brexit content.

Pages such as [Visit Europe after Brexit](https://www.gov.uk/email-signup/?topic=/government/visit-europe-after-brexit)
are sending people to 404 pages because the generated "Back" links go to
pages that don't exist because the taxon is in alpha stage.

This is not ideal, but we need to support this in the short term.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -8,4 +8,8 @@ module ApplicationHelper
   def title(text, params = {})
     render 'govuk_publishing_components/components/title', { title: text }.merge(params)
   end
+
+  def live_taxon?(taxon)
+    taxon['phase'] == 'live'
+  end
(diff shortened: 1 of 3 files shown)


karlbaker02:
Is it possible / should we be linking people to alpha taxons? Alternatively another quick fix might be to put this topic into beta (what does it mean for a taxon to be in either?)

sihugh:
Ta!  Yeah, it's a ~slight~ butchery of the taxonomy, but these are sub-sub-taxons created to hold one doc only, so having a taxon page for them doesn't feel right either.  Si is tracking these horrors so we can unravel them _soon_

PR #451, opened 2019-04-11
Enable users to signup to all content by organisation

bilbof:
This will enable users to subscribe to organisations content items via the /email-signup?link=ministry-of-magic endpoint.

It will behave in the same way as the taxonomy signup does currently.

Most of the changes are related to generalising the existing taxonomy endpoint to support all links, not just taxons.

I refactored the taxonomy code and then added in the organisations logic. It'll be easiest to review commit by commit.

Trello: https://trello.com/c/KsDrrchK/620.

Go to https://finder-frontend-pr-X.herokuapp.com/email-signup?link=hm-treasury to try it out.

## Views:

Screen Shot 2019-04-11 at 17 00 40
Screen Shot 2019-04-11 at 17 00 35


app/models/email_volume/organisation_weekly_email_volume.rb
```diff
@@ -0,0 +1,37 @@
+module EmailVolume
+  class OrganisationWeeklyEmailVolume
+    HIGH = '40 - 60'.freeze
+    MEDIUM = '0 - 20'.freeze
+    LOW = '0 - 5'.freeze
```
barrucadu: Are these numbers based on anything?

bilbof: It's very much a finger-in-air estimate. This is the same as our estimates for taxonomy emails (which may now be wrong).

This is the explanation in the commit (https://github.com/alphagov/email-alert-frontend/commit/d4a97b2b345c5f85f731d55075336e9ae70c6bdc#diff-5c90fde72d5dd7ed3acd972c5bc07f58) where the taxonomy estimate was added:

> These figures are based on analysis of historical email volume.

Perhaps we could make a request to email-alert-api to figure out the exact volume.

PR #461, opened 2019-05-02
Adds option for pre-selecting one of the frequency options

ChrisBAshton:
Trello: https://trello.com/c/f7DWeg4q/

## What
Adds a `default_frequency` parameter to the email alert frontend:

- If omitted (as it will be by default), nothing changes - the "immediately" option is selected as normal.
- If specified, this will override the default select option, i.e. `default_frequency=weekly` would set the selected frequency to weekly.

## Why
We want to set the default selection to 'daily' for the Business Finder.


ChrisBAshton:
Finder Frontend equivalent PR: https://github.com/alphagov/finder-frontend/pull/1104

PR #474, opened 2019-05-15
Smarter unsubscribe logic: match old subscriptions with new ones

ChrisBAshton:
Trello card: https://trello.com/c/VlnvU2Dl

## What & Why

- Updates the `UnsubscriptionController` to follow this logic:
  - If their latest subscription has ended already, show the "You're already unsubscribed" page.
  - Else if this is their latest subscription, render a "Are you sure you want to unsubscribe?" page.
  - Else, if this is not their latest subscription, but its frequency matches that of their latest subscription, then we can act on the user's intention and unsubscribe them.
  - Finally, if this is an old subscription and its frequency is different to that of their latest subscription, we cannot act on their instruction to unsubscribe as it may not be what the user intended. Instead we redirect them to manage their subscription manually.
- Updates version of `gds-api-adapter`, as we need the `get_latest_matching_subscription` query added in 59.4.
- Writes some integration tests.

## Note

I don't believe any tests or application code is required around authentication vs non-authentication. The actual unsubscription isn't being affected here - just the page that asks whether or not you confirm you want to unsubscribe. Clicking 'confirm' creates a POST method that _does_ rely on authentication.

### DIFF
--- a/app/controllers/unsubscriptions_controller.rb
+++ b/app/controllers/unsubscriptions_controller.rb
@@ -2,7 +2,15 @@ class UnsubscriptionsController < ApplicationController
   before_action :set_attributes
 
   def confirm
-    return render :confirm_already_unsubscribed if subscription_ended?
+    if subscription_ended?(@latest_subscription)
+      render :confirm_already_unsubscribed
+    elsif subscription_is_latest?
+      render :confirm
+    elsif old_subscription_has_same_frequency_as_latest?
+      redirect_to_unsubscribe_latest
+    else
+      redirect_to_manage_subscriptions
+    end
   end
 
   def confirmed
@@ -28,13 +36,30 @@ private
 
   def set_attributes
     @id = params.require(:id)
-    @subscription = api.get_subscription(@id)
-    @title = @subscription.dig("subscription", "subscriber_list", "title").presence
-    @authenticated_for_subscription = check_authenticated(@subscription)
+    @original_subscription = api.get_subscription(@id)
+    @latest_subscription = api.get_latest_matching_subscription(@id)
+    @title = @latest_subscription.dig("subscription", "subscriber_list", "title").presence
+    @authenticated_for_subscription = check_authenticated(@latest_subscription)
   end
 
-  def subscription_ended?
-    @subscription.dig("subscription", "ended_at").present?
+  def subscription_is_latest?
+    id(@original_subscription) == id(@latest_subscription)
+  end
+
+  def old_subscription_has_same_frequency_as_latest?
+    frequency(@latest_subscription) == frequency(@original_subscription)
+  end
+
+  def subscription_ended?(subscription)
+    subscription.dig("subscription", "ended_at").present?
+  end
+
+  def id(subscription)
+    subscription.dig("subscription", "id")
+  end
+
+  def frequency(subscription)
+    subscription.dig("subscription", "frequency")
   end
 
   def check_authenticated(subscription)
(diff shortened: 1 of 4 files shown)


app/controllers/unsubscriptions_controller.rb
```diff
@@ -28,13 +36,31 @@ def confirmed
...
+  def subscription_ended?(subscription)
+    subscription.dig("subscription", "ended_at").present?
+  end
+
+  def already_unsubscribed
+    subscription_ended(@latest_subscription)
+  end
+
+  def unsubscribe_current
+    # do nothing - yield to the confirm.html render
```
ChrisBAshton: This is a bit of a code smell. I really wanted this method to exist because it makes the `confirm` method logic really easy to read. My understanding is that Rails will yield to whatever view matches the current controller's public method name routed via `routes.rb`, provided you don't `return` anything in that method.

app/controllers/unsubscriptions_controller.rb
```diff
@@ -28,13 +36,36 @@ def confirmed
...
-    @subscription = api.get_subscription(@id)
-    @title = @subscription.dig("subscription", "subscriber_list", "title").presence
-    @authenticated_for_subscription = check_authenticated(@subscription)
+    @original_subscription = api.get_subscription(@id)
+    @latest_subscription = api.get_latest_matching_subscription(@id)
+    @title = @latest_subscription.dig("subscription", "subscriber_list", "title").presence
+    @authenticated_for_subscription = check_authenticated(@latest_subscription)
+  end
+
+  def subscription_is_latest
```
brucebolt: For this method, `old_subscription_has_same_frequency_as_latest` and `already_unsubscribed`, I'd suggest adding a question mark to the end of the name, since a boolean is returned

app/controllers/unsubscriptions_controller.rb
```diff
@@ -49,4 +80,17 @@ def check_authenticated(subscription)
...
+    redirect_to "/email/unsubscribe/#{id(@latest_subscription)}"
+  end
+
+  def redirect_to_manage_subscriptions
+    # Whilst it would be nice to direct to 'https://www.gov.uk/email/manage' instead
+    # (and treat the user as authenticated), this would open a security hole whereby
+    # if user A forwards an email to user B, then subsequently changes their
+    # frequency, user B could click on the old unsubscribe link and view user A's
+    # subscriptions. So DON'T change this line:
+    redirect_to 'https://www.gov.uk/email/authenticate'
```
brucebolt: In environments that aren't production, it won't always be `www.gov.uk` that we want to direct users to (e.g. in staging it would be `www.staging.publishing.service.gov.uk`.  Would it be possible to use Plek to determine the base for this redirect?  IIRC `Plek.new.website_uri` returns the correct value.

app/controllers/unsubscriptions_controller.rb
ChrisBAshton: Love it! Keep forgetting I can do that.

app/controllers/unsubscriptions_controller.rb
```diff
@@ -28,13 +36,34 @@ def confirmed
...
-    @subscription.dig("subscription", "ended_at").present?
+  def id(subscription)
+    subscription.dig("subscription", "id")
+  end
+
+  def frequency(subscription)
+    subscription.dig("subscription", "frequency")
+  end
+
+  def already_unsubscribed
```
brucebolt: I think this method should also have a question mark at the end of the name

ChrisBAshton: Actually it can be removed altogether! Good spot.

PR #498, opened 2019-06-04
Add combine_mode parameter

oscarwyatt:
We should find_or_create_subscriber_lists taking into account
whether a finder signup content item has a combine_mode
specified. 

Strictly speaking, this probably isn't necessary for the time being as the affected controller here (AFAIK) is only used for [topic page signups](https://www.gov.uk/education) but it can't hurt in case these changes stick around for longer than we expect

Trello: https://trello.com/c/y6ZhETsr/130-implement-a-joined-subscriber-list-model-in-email-alert-api-recommended-%F0%9F%8D%90

### DIFF
--- a/app/models/email_alert_signup.rb
+++ b/app/models/email_alert_signup.rb
@@ -73,6 +73,8 @@ private
       subscription_params[:links] = subscriber_list['links'].to_h
     end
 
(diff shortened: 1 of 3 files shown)


ChrisBAshton:
Closing as no longer needed.

PR #546, opened 2019-08-16
Add other document types to be permitted for signup

1pretz1:
This commit allows the following document types - ministerial_role,
person, topical_event and word_location to be initially signed up using
email-alert-frontend.

This is needed so when we retire the whitehall email signup pages we can
redirect here.

Trello:
https://trello.com/c/tgWUR1dx/25-switch-whitehalls-email-signup-pages-to-use-email-alert-frontends

### DIFF
--- a/app/models/content_item_subscriber_list.rb
+++ b/app/models/content_item_subscriber_list.rb
@@ -32,26 +32,64 @@ private
   def link_hash
     case content_item_type
     when 'taxon'
-      taxon_links
+      taxon_or_world_location_links
     when 'organisation'
       organisation_links
+    when 'person'
+      person_links
+    when 'ministerial_role'
+      ministerial_role_links
+    when 'topical_event'
+      topical_event_links
+    when 'world_location'
+      world_location_links
     else
       message = "No link hash available for content items of type #{content_item_type}!"
       raise UnsupportedContentItemError, message
     end
   end
 
-  def taxon_links
+  def taxon_or_world_location_links
+    if content_item['base_path'].match(%r{^/world/(.*)})
+      {
+        'world_locations' => [content_item['content_id']]
+      }
+    else
+      {
+        # 'taxon_tree' is the key used in email-alert-service for
+        # notifications, so create a subscriber list with this key.
+        'taxon_tree' => [content_item['content_id']]
+      }
+    end
+  end
+
+  def world_location_links
     {
-      # 'taxon_tree' is the key used in email-alert-service for
-      # notifications, so create a subscriber list with this key.
-      'taxon_tree' => [content_item['content_id']]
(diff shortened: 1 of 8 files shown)


app/models/content_item_subscriber_list.rb
```diff
@@ -32,20 +32,38 @@ def subscription_params
...
     else
       message = "No link hash available for content items of type #{content_item_type}!"
       raise UnsupportedContentItemError, message
     end
   end
 
-  def taxon_links
+  def taxon_or_world_location_links
+    if content_item['base_path'].match(%r{^/world/(.*)})
+      world_location_links
```
kevindew: I'd probably just put the world location links hash here since it's very short. Otherwise this seems to have an in-balance with the else condition.

1pretz1: Sure

1pretz1:
I don't think so @kevindew, as this controller only deals with specific content items and not finders, it states this at the top of the controller -
```
# This controller takes any content item and makes it possible to subscribe
# to alerts when documents linked to them are changed (taxons, orgs etc).
# This is in contrast to EmailAlertSignupsController, which takes a
# finder_email_signup content item.
```

This seems to be the reason why there is two signup endpoints for email-alert-frontend.

kevindew:
> I don't think so @kevindew, as this controller only deals with specific content items and not finders, it states this at the top of the controller -
> 
> ```
> # This controller takes any content item and makes it possible to subscribe
> # to alerts when documents linked to them are changed (taxons, orgs etc).
> # This is in contrast to EmailAlertSignupsController, which takes a
> # finder_email_signup content item.
> ```
> 
> This seems to be the reason why there is two signup endpoints for email-alert-frontend.

Sure, the bit I'm not sure about is the other bits that the Whitehall one does https://github.com/alphagov/whitehall/blob/31441ba6837aebeb9f1d9ffda4bd65e5b66ef0bb/lib/url_to_subscriber_list_criteria.rb#L58-L72 which seems to allow filtering of things by org, topics and subtaxons (what is a subtaxon?). This might all be moot though if they're covered by finder links.

1pretz1:
Yep it does allow those types of filtering which were originally for the old whitehall finders, this PR doesn't address this. 

These old finder email links shouldn't be available anywhere on gov.uk apart from for things like the old translated publications finder (https://www.gov.uk/government/publications.cy) however email subscriptions don't currently work for translated pages anyway.

So the only places these filtered url's will be accessed from is if google has indexed the page, I've addressed this partially in the whitehall pr linked above by redirecting the user to the corresponding new finder with all the filters pre-filled.

Also I have no idea what a subtaxon is or as it turns out a topic or a world location. 🤣

PR #551, opened 2019-08-29
Amend back button behaviour

1pretz1:
This addresses one signup edge case where if a user clicks through the
signup flow and lands on the address input page and then decides to
travel back using our back link, they will first arrive on the frequency
selection page then if they click back again they will be redirected to
gov.uk's homepage, this is obviously not a great user experiance and
affects any email signup flow.

It is also possible to get trapped in an infinate loop using the back
links between the confirm page and the frequency page for signups that
start in email-alert-frontend.

This commit targets one edge case regarding email subscriptions from
/get-ready-brexit-check/results as we now can access the subscriber list url to
find the original page the user signed up to. This is hacky and a solution
to fix the overriding problem will need to be thought of.


app/controllers/subscriptions_controller.rb
```diff
@@ -79,6 +79,8 @@ def govuk_url
       sanitised_referer_uri.query = referer_uri.query
       sanitised_referer_uri.fragment = referer_uri.fragment
       sanitised_referer_uri.to_s
+    elsif @subscriber_list["url"].match?(%r{^/get-ready-brexit-check/results})
```
thomasleese: I'm wondering if we even need this check because right now only subscriber lists from the tool will have such URLs (and in the future we might want to support any subscriber list with a URL supporting the back button)... but not a big deal if we want to get this out.

1pretz1: Yup true, when we think of an actual solution to the main problem this can be edited accordingly.

1pretz1: And having the original page url on all subscriber lists would improve the journey but still isn't great as really the back button should go to the previous page and not shoot the user to the original page which would skip a few pages in the signup flow, like the confirm and subscribe pages for example.

PR #608, opened 2019-11-22
Deprecate top-level '/authenticate' routes

benthorner:

This moves the existing '/authenticate' routes under the '/manage' scope to
reflect that they are specific to the actions within and support the
upcoming work that will introduce additional authentication routes for a
new subscription (as opposed to managing subscriptions / loggin in).

Please see the commits for more details.


benthorner:
@kevindew I think you'd be a good person to review this, as it sets the direction for the rest of our work with the frontend.

config/routes.rb
```diff
@@ -31,9 +35,9 @@
       get "/complete" => "subscriptions#complete", as: :subscription
     end
 
-    get "/authenticate" => "authentication#sign_in", as: :sign_in
-    post "/authenticate" => "authentication#request_sign_in_token", as: :request_sign_in_token
-    get "/authenticate/process" => "authentication#process_sign_in_token", as: :process_sign_in_token
+    # DEPRECATED: legacy routes for managing subscriptions
+    get "/authenticate" => "subscriber_authentication#sign_in"
```
kevindew: Based on our IRL chat it sounds like we'll need to keep this route long term so it might be better as a redirect.

config/routes.rb
```diff
@@ -31,9 +35,9 @@
...
+    get "/authenticate/process" => "subscriber_authentication#process_sign_in_token"
```
kevindew: It might be good to flag this with the date we can delete it from if this will soon be safe to remove.

benthorner: Updated - how does it look now?

PR #609, opened 2019-11-25
[DO NOT MERGE] Remove legacy route to process sign in token

benthorner:

Previously we scoped this route under '/manage'. This removes the
old route, which we kept for backwards compatibility. New auth emails
should now be using the scoped URL, and since after 7 days all the
emails containing the old URL are now expired, we think it's reasonable
to completely remove the old route.


benthorner:
Confirmed this is no longer in use...

Screenshot 2019-12-17 at 15 42 46

PR #615, opened 2019-11-26
Add new opt-in subscribe feature

benthorner:

This adds a new feature to create (or update) a subscription against a
validated JWT token. Since the address to subscribe to is encrypted within
the token, we can be sure the request represents genuine intent. As an
additional protection, we also validate the subscriber list matches the one
in the token, so that it can't be used out of context.

### DIFF
--- a/app/controllers/subscriber_authentication_controller.rb
+++ b/app/controllers/subscriber_authentication_controller.rb
@@ -30,30 +30,21 @@ class SubscriberAuthenticationController < ApplicationController
   end
 
   def process_sign_in_token
-    subscriber_id, redirect = read_token(params.require(:token))
-
-    if subscriber_id.nil?
+    unless token.valid?
       deauthenticate_subscriber
       flash[:error_summary] = "bad_token"
       return redirect_to :sign_in
     end
 
-    authenticate_subscriber(subscriber_id)
-
-    destination = safe_redirect_destination(redirect) || list_subscriptions_path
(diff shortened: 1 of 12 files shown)


app/controllers/subscriber_authentication_controller.rb
```diff
@@ -30,30 +30,21 @@ def request_sign_in_token
   end
 
   def process_sign_in_token
-    subscriber_id, redirect = read_token(params.require(:token))
-
-    if subscriber_id.nil?
+    unless token.valid?
```
huwd: how come this doesn't need to be an instance variable `@token`?

huwd: urgh never mind, didn't see line 46 when i wrote this

app/controllers/subscription_authentication_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+
+private
+
+  def valid_params?
+    token.data[:topic_id] == @topic_id &&
+      valid_frequencies.include?(@frequency)
+  end
+
+  def token
+    @token ||= AuthToken.new(params.require(:token))
```
koetsier: Could move this to line 7 so all the params are read in one place

benthorner: Almost: I think we need to keep it here, to avoid passing it into `valid_params?` manually.

PR #624, opened 2019-12-03
Stop preselecting email frequency when subscribing

vanitabarrett:
Trello: https://trello.com/c/KYrvoZmz/306-double-opt-in-stop-preselecting-the-frequency-when-subscribing
Smokey PR to remove associated tests: https://github.com/alphagov/smokey/pull/622

## What
When users sign up for a new email subscription, the email frequency is normally pre-selected. This PR removes that functionality so no radio button is preselected and users are required to select an option before they move on to the next step.

The update subscription frequency behaviour remains the same, as it pre-selects the user's **current** chosen frequency rather than a default.

## New Subscription (before)
Screenshot 2019-12-04 at 15 19 30

## New Subscription (after)
Screenshot 2019-12-04 at 15 19 44

## New subscription error
Screenshot 2019-12-04 at 14 38 11


app/helpers/frequencies_helper.rb
```diff
@@ -3,13 +3,15 @@ def valid_frequencies
     t("frequencies").map { |frequency, _config| frequency.to_s }
   end
 
-  def frequencies(options)
+  def frequencies(options = false)
```
benthorner: Could you help me understand this change? It's a bit strange to default an argument to a value of a different type (`options` would be a hash). Perhaps we can pass in `checked_frequency => nil` instead?

app/views/subscriptions/new_frequency.html.erb
```diff
@@ -9,17 +9,31 @@
...
+      } %>
+    <% end %>
+
     <h1 class="govuk-heading-l">Get emails when pages are added or updated</h1>
 
     <%= form_tag subscription_frequency_path, method: :post, class: "checklist-email-signup" do %>
       <%= hidden_field_tag :topic_id, @topic_id %>
 
       <%= render "govuk_publishing_components/components/fieldset", {
+        id: "email-frequency-input",
```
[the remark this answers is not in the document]
vanitabarrett: Yes - this is used by the [error message above](https://github.com/alphagov/email-alert-frontend/pull/624/files/4b455e1a1b2e2b7eda4c1b6dc893046f48204c72#diff-e7f735aec01a0f80a6dc61aa6b4f0d9bR18) to link to the question that is erroring. The [component guide](https://components.publishing.service.gov.uk/component-guide/error_summary) shows an example of how this component works.

spec/features/subscribe_default_frequency_spec.rb
```diff
@@ -7,7 +7,7 @@
 
   scenario do
     when_i_signup_with_a_default_frequency
-    then_i_see_the_frequency_is_preset
+    then_i_see_the_frequency_is_not_preset
```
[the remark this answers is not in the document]
vanitabarrett: True! It helped me check that this new functionality is working as expected as I was working on it, but happy to delete now.

app/helpers/frequencies_helper.rb
vanitabarrett: That sounds like a more consistent option, I'll do that 👍

app/views/subscriptions/new_frequency.html.erb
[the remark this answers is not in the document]
benthorner: Oh yeah (duh!). I used to know this. Thanks for explaining.

app/helpers/frequencies_helper.rb
benthorner: We should be able to simplify this even further: using the original helper, but defaulting `options` to an empty hash, and passing nothing in the `new_frequency` view. 

The modified helper code works, but it's hard to understand because it no longer sets `checked` to `true` or `false` - now it's either a string or `nil`.

vanitabarrett: D'oh. I totally misread how this worked the first time. Will fix now.

vanitabarrett: @benthorner Changed - I think that should be ok now?

app/views/subscriptions/new_frequency.html.erb
```diff
@@ -9,17 +9,31 @@
...
         legend_text: '<h2 class="govuk-heading-m">How often do you want to get updates?</h2>'.html_safe,
+        error_message: flash[:error],
       } do %>
         <%= render "govuk_publishing_components/components/radio", {
           name: "frequency",
-          items: frequencies(:checked_frequency => @default_frequency)
+          items: frequencies(),
```
benthorner: Very minor: the parentheses aren't necessary here :-).

PR #629, opened 2019-12-05
Send verification email before subscribing

benthorner:
https://trello.com/c/dBwLRlDm/283-double-opt-in-send-double-opt-in-email-before-creating-subscription

This replaces the current behaviour of the email subscribe journey by
sending a verification email, which the user must receive to continue. In
order to avoid having the user's email address in the URL, the view for
checking the email is rendered as part of the 'create' action.

Depends on: https://github.com/alphagov/gds-api-adapters/pull/965


app/controllers/subscription_authentication_controller.rb
```diff
@@ -25,7 +25,16 @@ def authenticate
...
+    redirect_to subscription_complete_path(topic_id: @topic_id, frequency: @frequency)
+  end
+
+  def complete
+    topic_id = params.require(:topic_id)
+    @frequency = params.require(:frequency)
+
+    @title = email_alert_api
+      .get_subscriber_list(slug: topic_id)
+      .to_h.dig("subscriber_list", "title")
```
koetsier: I don't think the 'to_h' is strictly necessary

benthorner: Unfortunately we do: turns out it's a `GdsApi::Response`, which doesn't support `fetch` or `dig` directly :-(.

config/locales/subscriptions.yml
```diff
@@ -4,17 +4,24 @@ en:
...
-    complete:
-      summary:
-        immediately: "You’ll get an email each time there’s an update to ‘%{title}’"
-        daily: "You’ll get an email each day there’s been an update to ‘%{title}’"
-        weekly: "You’ll get one email per week if there’s been an update to ‘%{title}’"
+    check_email:
+      title: Check your email inbox to confirm your subscription
+      description_html: "We’ve sent an email to: <strong>‘%{address}’</strong>"
+      instruction: Check your email and confirm your email address.
+      expiry: You have 7 days to confirm
```
koetsier: add full stop You have 7 days to confirm.

config/locales/subscriptions.yml
```diff
@@ -4,17 +4,24 @@ en:
...
         weekly: "You'll get one email per week if a page is added or changed."
-    complete:
-      summary:
-        immediately: "You’ll get an email each time there’s an update to ‘%{title}’"
-        daily: "You’ll get an email each day there’s been an update to ‘%{title}’"
-        weekly: "You’ll get one email per week if there’s been an update to ‘%{title}’"
+    check_email:
+      title: Check your email inbox to confirm your subscription
+      description_html: "We’ve sent an email to: <strong>‘%{address}’</strong>"
+      instruction: Check your email and confirm your email address.
```
koetsier: instruction: Use the link in the email to confirm your subscription.

PR #641, opened 2019-12-18
Several UI improvements

koetsier:
Some last minute improvements to the UI.

https://trello.com/c/umMOCXZa/328-double-opt-in-implement-last-few-cosmetic-content-updates

### DIFF
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -1,9 +1,9 @@
 class SubscriptionsController < ApplicationController
   include FrequenciesHelper
   before_action :assign_attributes
-  before_action :assign_back_url
 
   def new
+    @back_url = @frequency.blank? ? govuk_url : url_for(action: :new, topic_id: @topic_id)
     if @frequency.present?
       return frequency_form_redirect unless valid_frequency
 
@@ -23,6 +23,7 @@ class SubscriptionsController < ApplicationController
       )
     else
       flash.now[:error] = t("subscriptions.new_frequency.missing_frequency")
+      @back_url = url_for(action: :new, topic_id: @topic_id)
       render :new_frequency
     end
   end
@@ -30,6 +31,11 @@ class SubscriptionsController < ApplicationController
   def verify
     return frequency_form_redirect unless valid_frequency
 
+    @back_url = url_for(host: Plek.new.website_root,
+                        action: :new,
+                        topic_id: @topic_id,
+                        frequency: @frequency,
+                        address: @address)
     if @address.present? && subscribe
       render :check_email
     else
@@ -52,11 +58,6 @@ private
     @frequency = subscription_params[:frequency]
     @address = subscription_params[:address]
     @title = @subscriber_list["title"]
-  end
-
-  def assign_back_url
-    @back_url = url_for(action: :new, topic_id: @topic_id)
-    @back_url = govuk_url if params[:action] == "new" && @frequency.blank?
   end
 
   def subscription_params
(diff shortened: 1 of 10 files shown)


app/views/subscriptions/check_email.html.erb
```diff
@@ -1,3 +1,9 @@
+<% content_for :back_link do %>
+  <%= render "govuk_publishing_components/components/back_link", {
+      href: :back
```
benthorner: Should this be `@back_url`?

config/locales/subscription_authentication.yml
```diff
@@ -7,6 +7,6 @@ en:
       browser_title: "Subscription created successfully"
       title: "You’ve subscribed successfully"
       summary:
-        immediately: "You’ll get an email each time there’s an update to ‘%{title}’"
-        daily: "You’ll get an email each day there’s been an update to ‘%{title}’"
-        weekly: "You’ll get one email per week if there’s been an update to ‘%{title}’"
+        immediately: "You’ll get an email each time there’s an update to <em>%{title}</em>"
```
koetsier: should be bold instead of italic

koetsier: remove the em-s

app/views/subscriptions/check_email.html.erb
koetsier: updated that and added a new ```@back_url``` definition

app/controllers/subscriptions_controller.rb
```diff
@@ -1,9 +1,9 @@
 class SubscriptionsController < ApplicationController
   include FrequenciesHelper
   before_action :assign_attributes
-  before_action :assign_back_url
 
   def new
+    assign_back_url
```
benthorner: This is better than doing it everywhere! I'm finding the change a bit confusing, though, as here we're calling `assign_back_url`, whereas on line 34, we're assigning `@back_url` manually. Looking at the `assign_back_url`, I can see that each line is specific to one of the two places where it's called.

For simplicity and consistency, could we remove the `assign_back_url` method altogether?

koetsier: good point. fixed

spec/features/subscribe_back_spec.rb
```diff
@@ -27,10 +34,29 @@ def given_there_is_content_i_can_subscribe_to
...
+  end
+
   def when_i_start_signup_via_govuk
     visit "/email/subscriptions/new?topic_id=#{@topic_id}"
   end
 
+  def and_i_select_frequency_and_submit
+    stub_email_alert_api_sends_subscription_verification_email("person@somewhere.uk", "daily", @topic_id)
+    choose(option: "daily")
+    within "form.checklist-email-signup" do
```
benthorner: Minor: could we just do `click_on "Continue"` or something? I guess this approach might be more robust, but on the other hand, I had to think more about what the code is doing.

spec/features/subscribe_back_spec.rb
```diff
@@ -15,6 +15,13 @@
     then_i_can_only_go_back_to_govuk
   end
 
+  scenario "confirm email" do
+    when_i_start_signup_via_govuk
+    and_i_select_frequency_and_submit
+    and_i_fill_in_my_email
+    then_i_can_see_a_back_url_with_correct_parameters
```
benthorner: Minor: it's a bit strange to read this in a (UI) feature test, since the user can't actually "see a back URL". How about "then_i_can_go_back_to_the_frequency_page" or similar?

PR #654, opened 2019-12-30
Add feature tests for managing subscriptions

emmabeynon:
https://trello.com/c/wb8hyN7U/308-double-opt-in-add-feature-tests-for-managing-subscriptions

### DIFF
--- a/spec/features/change_email_frequency_spec.rb
+++ b/spec/features/change_email_frequency_spec.rb
@@ -1,0 +1,80 @@
+RSpec.feature "Change email frequency after receiving confirmation link" do
+  include GdsApi::TestHelpers::EmailAlertApi
+  include TokenHelper
+
+  scenario do
+    given_i_have_a_subscription
+    when_i_visit_the_manage_my_subscriptions_page
+    then_i_can_see_i_am_subscribed_to_daily_updates
+    when_i_click_to_change_how_often_i_get_updates
+    and_i_select_once_a_week
+    then_i_can_see_that_i_am_subscribed_to_weekly_updates
+  end
+
+  def given_i_have_a_subscription
+    @address = "test@test.com"
+    @subscriber_id = SecureRandom.uuid
+    @subscription_id = SecureRandom.uuid
+    @subscription_title = "Something to subscribe to"
+
+    stub_email_alert_api_has_subscriber_subscriptions(
+      @subscriber_id, @address, nil, subscriptions: [subscription]
+    ).then
+     .to_return(status: 200,
+                body: subscriber_subscriptions_response_with_weekly_frequency)
+  end
+
+  def when_i_visit_the_manage_my_subscriptions_page
+    token = jwt_token(data: {
+      "address" => @address,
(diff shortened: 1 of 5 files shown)


spec/features/change_email_frequency_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  def given_i_have_a_subscription
+    @address = "test@test.com"
+    @subscriber_id = SecureRandom.uuid
+    @subscription_id = SecureRandom.uuid
+    @subscription_title = "Something to subscribe to"
+
+    stub_email_alert_api_has_subscriber_subscriptions(
+      @subscriber_id, @address, nil, subscriptions: [subscription]
+    ).times(3)
```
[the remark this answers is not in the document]
benthorner: Ah, I think it's for the flash at the end, but I don't think it's required. Does it work without?

spec/features/unsubscribe_all_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    given_i_have_multiple_subscriptions
+    when_i_visit_the_manage_my_subscriptions_page
+    and_i_click_on_unsubscribe_from_everything
+    and_i_confirm_to_unsubscribe
+    then_i_can_see_i_have_been_unsubscribed
+  end
+
+  def given_i_have_multiple_subscriptions
+    @address = "test@test.com"
+    @subscriber_id = 123
```
benthorner: Minor: I notice we use `SecureRandom` in the other specs. Any reason to hard code it?

spec/features/unsubscribe_all_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    stub_email_alert_api_has_subscriber_subscriptions(
+      @subscriber_id,
+      @address,
+      nil,
+      subscriptions: [subscription, subscription],
+    ).times(3)
+     .then
+     .to_return(status: 200, body: subscriber_with_no_subscriptions_response)
```
benthorner: Since the `times(3)` is a bit tricky to understand, can we simplify to just return the original response each time? I understand it's not as realistic, but I don't think anyone will notice.

spec/features/change_email_frequency_spec.rb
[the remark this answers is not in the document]
emmabeynon: Looks like it does work without - I was following the WebMock docs which suggested this pattern but since we don't need it, I've removed 👍

spec/features/unsubscribe_all_spec.rb
emmabeynon: No, just an oversight, I've changed to use `SecureRandom`.

spec/features/unsubscribe_all_spec.rb
emmabeynon: Sure, I've removed it 👍

emmabeynon:
Thanks @benthorner - all comments should be addressed now, and I've got another PR on the way to address the translations.

PR #655, opened 2019-12-31
Extract translations for subscriptions management and subscriber authentication

emmabeynon:
For https://trello.com/c/wb8hyN7U/308-double-opt-in-add-feature-tests-for-managing-subscriptions

### DIFF
--- a/app/views/subscriptions_management/confirm_unsubscribe_all.html.erb
+++ b/app/views/subscriptions_management/confirm_unsubscribe_all.html.erb
@@ -1,4 +1,4 @@
-<% content_for :title, 'Are you sure you want to unsubscribe from everything?' %>
+<% content_for :title, t("subscriptions_management.confirm_unsubscribe_all.heading") %>
 
 <% content_for :back_link do %>
   <%= render "govuk_publishing_components/components/back_link", {
@@ -8,9 +8,9 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h1 class="govuk-heading-l">Are you sure you want to unsubscribe from everything?</h1>
+    <h1 class="govuk-heading-l"><%= t("subscriptions_management.confirm_unsubscribe_all.heading") %></h1>
 
-    <p class="govuk-body">You won’t get any more automated emails from GOV.UK.</p>
+    <p class="govuk-body"><%= t("subscriptions_management.confirm_unsubscribe_all.description") %></p>
 
     <%= form_tag(action: :confirmed_unsubscribe_all) do %>
       <%= hidden_field_tag(:from, @from) %>
--- a/app/views/unsubscriptions/confirm_already_unsubscribed.html.erb
+++ b/app/views/unsubscriptions/confirm_already_unsubscribed.html.erb
@@ -1,5 +1,5 @@
 <%
-page_title = "You’ve already unsubscribed"
+page_title = t("unsubscriptions.title.confirm_already_unsubscribed")
 page_title += " from #{@title}" if @title
 content_for :title, page_title
 %>
@@ -16,10 +16,6 @@ content_for :title, page_title
   <div class="govuk-grid-column-two-thirds">
     <h1 class="govuk-heading-l"><%= page_title %></h1>
 
-    <% if @title %>
-      <p class="govuk-body">You won’t get any more updates about <%= @title %>.</p>
-    <% else %>
-      <p class="govuk-body">You won’t get any more updates about this topic.</p>
-    <% end %>
+    <%= render "confirmation" %>
   </div>
 </div>
--- a/config/locales/subscriptions_management.yml
+++ b/config/locales/subscriptions_management.yml
@@ -1,10 +1,28 @@
 en:
   subscriptions_management:
+    heading: Manage your subscriptions
     index:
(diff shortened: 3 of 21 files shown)


spec/features/manage_subscriptions_confirmation_email_spec.rb
```diff
@@ -28,6 +28,9 @@ def and_i_enter_my_email_address
 
   def then_i_can_see_a_confirmation_email_has_been_sent_to_me
     expect(@request).to have_been_requested
-    expect(page).to have_content("We’ve sent an email to #{@email_address}")
+    expect(page).to have_content(
+      I18n.t("subscriber_authentication.request_sign_in_token.confirmation",
```
benthorner: Minor: we should use `t!` here to be sure this will fail with an exception if the translation doesn't exist.

app/views/subscriptions_management/confirm_unsubscribe_all.html.erb
```diff
@@ -1,4 +1,4 @@
-<% content_for :title, 'Are you sure you want to unsubscribe from everything?' %>
+<% content_for :title, t("subscriptions_management.confirm_unsubscribe_all.heading") %>
```
benthorner: Minor: I'm curious about the use of `heading`. In the past, I think we've used `title` everywhere. Doesn't really matter - just wondering if there's a reason for the choice of word?

app/views/subscriptions_management/confirm_unsubscribe_all.html.erb
```diff
@@ -8,11 +8,11 @@
...
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h1 class="govuk-heading-l">Are you sure you want to unsubscribe from everything?</h1>
+    <h1 class="govuk-heading-l"><%= t("subscriptions_management.confirm_unsubscribe_all.heading") %></h1>
 
-    <p class="govuk-body">You won’t get any more automated emails from GOV.UK.</p>
+    <p class="govuk-body"><%= t("subscriptions_management.confirm_unsubscribe_all.description") %></p>
 
-    <%= form_tag(action: :confirmed_unsubscribe_all) do %>
+    <%= form_tag(action: :confirm_unsubscribe_all) do %>
```
benthorner: Good spot - we probably don't need it at all, since it seemed to work without!

config/locales/subscriptions_management.yml
```diff
@@ -1,10 +1,28 @@
...
     index:
       subscription:
         immediately: "You chose to get updates as soon as they happen."
         daily: "You chose to get daily updates."
         weekly: "You chose to get weekly updates."
     update_address:
+      heading: Change your email address
+      current_email: "Your current email address is %{address}"
+      description: |
+        "If you want to transfer your subscription to %{address},
```
benthorner: I think we need to drop the quotes here, as otherwise they'll be part of the printed text.

app/views/subscriptions_management/confirm_unsubscribe_all.html.erb
emmabeynon: I think it's just the first place I extracted a translation was in a heading, and then spotted the same text was being used in the title, so just reused the translation I had already created.

config/locales/subscriptions_management.yml
emmabeynon: Done

spec/features/manage_subscriptions_confirmation_email_spec.rb
emmabeynon: Done

emmabeynon:
Thanks @benthorner - I've added translations to those tests and extracted translations for the unsubscription flows.

config/locales/unsubscriptions.yml
```diff
@@ -0,0 +1,9 @@
+en:
+  unsubscriptions:
+    title:
+      confirm_already_unsubscribed: You’ve already unsubscribed
+      confirm: Are you sure you want to unsubscribe?
+      confirmed: You’ve successfully unsubscribed
+    confirmation:
+      with_title: You won’t get any more updates about %{title}.
```
benthorner: Interesting - I thought you had to put quotes around translations with `%`, but apparently it works fine.

spec/controllers/unsubscriptions_controller_spec.rb
```diff
@@ -36,7 +36,9 @@
     it "renders the title on the page" do
       get :confirm, params: { id: id }
 
-      expect(response.body).to include("You won’t get any more updates about #{title}")
+      expect(response.body).to include(
+        I18n.t!("unsubscriptions.confirmation.with_title", title: title),
```
benthorner: Good spot!

app/views/unsubscriptions/confirm_already_unsubscribed.html.erb
```diff
@@ -1,5 +1,5 @@
 <%
-page_title = "You’ve already unsubscribed"
+page_title = t("unsubscriptions.title.confirm_already_unsubscribed")
```
benthorner: Minor: I think we usually do these by view i.e. `unsubscriptions.confirm_already_unsubscribed.title`.

PR #669, opened 2020-01-23
Follow redirects for content item email signups

koetsier:
Sometimes taxa are moved and a redirect is set in place.  Email signup links to the moved taxons are often hardcoded in content and will have to be updated. Some may be missed resulting
in broken links.

Previously, the case of redirecting /government/brexit -> /brexit was handled
as a special hard coded case.

This PR makes makes this more general by following the redirects in
the content item.

trello: https://trello.com/c/zhg0rOFS/386-update-government-brexit-redirect-in-email-alert-frontend

### DIFF
--- a/app/controllers/content_item_signups_controller.rb
+++ b/app/controllers/content_item_signups_controller.rb
@@ -5,6 +5,7 @@
 class ContentItemSignupsController < ApplicationController
   protect_from_forgery except: [:create]
   before_action :require_content_item_param
+  before_action :handle_redirects
   before_action :validate_document_type
   helper_method :estimated_email_frequency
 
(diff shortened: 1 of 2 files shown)


app/controllers/content_item_signups_controller.rb
```diff
@@ -59,6 +55,18 @@ def content_item
       .content_item(content_item_path)
   end
 
+  def handle_redirects
+    if content_item["document_type"] == "redirect"
+      destination_path = content_item.dig("redirects", 0, "destination")
```
huwd: Does this mean we added something to the document_type too then?

The previous one redirected to based on the topic?

PR #713, opened 2020-03-16
Make it clear that the email is to manage subscriptions

barrucadu:
We have some evidence that people are confused by having to "confirm" their email address when they already had to do that to sign up in the first place.

### DIFF
--- a/config/locales/subscriber_authentication.yml
+++ b/config/locales/subscriber_authentication.yml
@@ -2,7 +2,7 @@ en:
   subscriber_authentication:
     heading: Manage your subscriptions
     sign_in:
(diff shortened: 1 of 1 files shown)


config/locales/subscriber_authentication.yml
```diff
@@ -2,7 +2,7 @@ en:
   subscriber_authentication:
     heading: Manage your subscriptions
     sign_in:
-      description: You’ll need to confirm your email address before you can manage your subscriptions.  There may be a delay of up to 15 minutes before you receive the confirmation email.
+      description: We'll send you an email with a link in to manage your subscriptions.  There may be a delay of up to 15 minutes before you receive this email.
```
issyl0:
```suggestion
      description: We‘ll send you an email with a link in to manage your subscriptions.  There may be a delay of up to 15 minutes before you receive this email.
```

PR #715, opened 2020-03-16
Replace Rails default error pages with GOV.UK styled ones

benjamineskola:
This is primarily in order to reword the 422 error message, which is particularly unhelpful.

https://trello.com/c/eeMr0zaf/864-content-changes-to-email-unsubscribe-page


kevindew:
Hey - just saw this PR while I was looking for others in the repo. 

I'm afraid that only the 422 response will have any affect here. This is because our [nginx configuration intercepts error responses](https://github.com/alphagov/govuk-puppet/blob/0d43abfda781190f9a2fc00e49583ddc50d8f9a3/modules/nginx/templates/etc/nginx/nginx.conf.erb#L33) and [converts them](https://github.com/alphagov/govuk-puppet/blob/0d43abfda781190f9a2fc00e49583ddc50d8f9a3/modules/router/templates/router_include.conf.erb#L81-L88) into [pre-defined error pages](https://github.com/alphagov/static/tree/master/app/views/root). So you might find setting up a rails error app ends up being somewhat misleading and confusing (which is a shame because I do like them).

From this being flagged as "the change you wanted was rejected" do you know if users are seeing this due to the same exception or a variety? From a quick look it looked like there is this one (a lot): ActionController::InvalidAuthenticityToken. So we could decide that we want to handle this by having a rescue_from in ActionController

I also wonder why we see so many and if this may indicate some sort of session problem 🤔

benjamineskola:
The 422 response was the one we're specifically trying to handle, I included the others for completeness but if they can never work then I can remove them. (Actually, I seemed to get an nginx error for 422 as well when running in docker, but that may have been due to misunderstanding.)

This was meant to be a quick fix in order to give a clearer message to users, but if the underlying problem can be fixed more easily then that's even better.

kevindew:
@benjamineskola Sure, yeah we don't have the nginx configuration in docker to override error pages.

If we continue with a rescue response app you'll probably have to change the routes from get to match as per: https://github.com/alphagov/content-publisher/blob/a66058756fda192320d09d13e1f3b40af05594dc/config/routes.rb#L119-L125

vanitabarrett:
I've added a card to the Coronavirus Notifications team backlog to take a look at this and double check there's no weirdness going on here that we're not aware of. 
We'll prioritise it so it gets looked at next week. https://trello.com/c/mKFZC0nm/45-investigate-cause-of-your-change-was-rejected-error-message-on-authenticate

app/views/errors/error_422.html.erb
```diff
@@ -0,0 +1,6 @@
+<% content_for(:title, 'The change you wanted was rejected') %>
+
+<h1 class="page-title">
+  <%= @custom_header || 'The change you wanted was rejected' %>
+</h1>
+<p><%= @custom_body || 'Your link has expired, or your email does not exist in our database' %></p>
```
theseanything, line 6: I think user's are seeing this after clicking the confirm button on `/email/manage/authenticate` and `/email/subscriptions/new` page, not after clicking a link in their email so this error message is potentially confusing.

We catch and manage errors from missing email addresses, so I don't think user will see this page because a that.
Think this page is shown due to invalid CSFR tokens, either old cookies persisting or CSRF token in the HTML might be cached.

theseanything:
Superseded by #743

PR #735, opened 2020-03-31
Add coronavirus weekly email volume

laurentqro:
This sets expectation for the weekly volume of changes for the coronavirus topic.

Screenshot 2020-04-02 at 12 29 18

The implementation is somewhat hacky. However, the plan is to migrate the coronavirus topical event to a taxon, which means the hack will be removed then.

https://trello.com/c/1TbLyQXY/73-display-estimated-weekly-email-volume-for-coronavirus-email-sign-up

### DIFF
--- a/app/models/email_volume/taxon_weekly_email_volume.rb
+++ b/app/models/email_volume/taxon_weekly_email_volume.rb
@@ -1,8 +1,8 @@
 module EmailVolume
   class TaxonWeeklyEmailVolume
-    HIGH = "40 - 60".freeze
-    MEDIUM = "0 - 20".freeze
-    LOW = "0 - 5".freeze
+    HIGH = "40 to 60".freeze
+    MEDIUM = "0 to 20".freeze
+    LOW = "0 to 5".freeze
 
     def initialize(taxon)
       @taxon = taxon
--- a/spec/models/weekly_email_volume_spec.rb
+++ b/spec/models/weekly_email_volume_spec.rb
@@ -84,5 +84,24 @@ RSpec.describe EmailVolume::WeeklyEmailVolume do
         expect(described_class.new(person).estimate).to be_nil
       end
     end
+
+    context "given a topical event page" do
+      it "returns the estimated volume for the coronavirus page" do
+        coronavirus_page = {
+          document_type: "topical_event",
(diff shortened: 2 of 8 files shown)


vanitabarrett:
This looks ok to me, but I'd prefer to let @theseanything / @koetsier take a look as well before approving 🙂

spec/models/weekly_email_volume_spec.rb
```diff
@@ -84,5 +84,16 @@
...
+    context "given the coronavirus page" do
+      it "returns the estimated volume" do
+        coronavirus_page = {
+          document_type: "topical_event",
+          base_path: "/government/topical-events/coronavirus-covid-19-uk-government-response",
+        }.deep_stringify_keys
+
+        expect(described_class.new(coronavirus_page).estimate).to eql "40 to 60"
+      end
+    end
```
theseanything, line 97: Should we add a test when its not the coronavirus page that we aren't showing an estimate?

app/models/email_volume/taxon_weekly_email_volume.rb
```diff
@@ -1,8 +1,8 @@
 module EmailVolume
   class TaxonWeeklyEmailVolume
-    HIGH = "40 - 60".freeze
-    MEDIUM = "0 - 20".freeze
-    LOW = "0 - 5".freeze
+    HIGH = "40 to 60".freeze
+    MEDIUM = "0 to 20".freeze
+    LOW = "0 to 5".freeze
```
theseanything, line 5: Nit: Could these be moved into the module? Then we could share these constants between `TaxonWeeklyEmailVolume` and `OrganisationWeeklyEmailVolume`?

laurentqro: @theseanything  Yeah it did cross my mind to reduce duplication here, but I ended up not, because it's accidental duplication, i.e. the values _happen_ to be the same, but there's no fundamental reason (that I know of) why taxon and organisation email volumes would always be identical.

Happy to hear your thoughts on this.

theseanything: Ah yeah, I guess we could see different volumes, your right! Probably safer to leave, them so its easier to tweak later.

PR #743, opened 2020-04-03
Add error page for invalid sessions

theseanything:
Previously we were showing users the default rails 422 pages on InvalidAuthenticityToken exceptions. This adds a GOV.UK styled error page, thats should be less confusing and tell the users to try the form again.

## Before
Screenshot 2020-04-08 at 15 30 35

## After
Screenshot 2020-04-09 at 11 05 10

### DIFF
--- a/spec/controllers/application_controller_spec.rb
+++ b/spec/controllers/application_controller_spec.rb
@@ -1,0 +1,16 @@
+RSpec.describe ApplicationController do
+  render_views
+  controller do
+    def index
+      raise ActionController::InvalidAuthenticityToken
+    end
+  end
+
+  describe "handling InvalidAuthenticityToken exceptions" do
+    it "renders the invalid session template" do
+      get :index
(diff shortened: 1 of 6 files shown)


spec/controllers/application_controller_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+  controller do
+    def index
+      raise ActionController::InvalidAuthenticityToken
+    end
+  end
+
+  describe "handling InvalidAuthenticityToken exceptions" do
+    it "renders the invalid session template" do
+      get :index
+      expect(response.body).to include(I18n.t!("sessions.invalid.heading"))
```
laurentqro, line 12: Gonna be a bit nitpicky here. :)

If someone changes the value associated with `sessions.invalid.heading` to be something wrong, then this test would not pick up on it.

Also, it kind of knows too much about the internals of what it's testing. It would break if we ever decided to change the location of the text, e.g. `sessions.errors.invalid.heading`.

PR #746, opened 2020-04-07
Override topical event title for email

vanitabarrett:
Trello: https://trello.com/c/x2QeWfip/61-rename-coronavirus-email-subscription-title

## What
Temporarily override the Coronavirus topical event title for the email signup journey.

## Why
We added a [migration task to rename the email subscriber list title itself](https://github.com/alphagov/email-alert-api/pull/1199), but this isn't used in the email alert frontend when someone is signing up. Instead, this relies on the content_item of (in this case) the topical event. 

We could rename the topical event itself in Whitehall but this is may need more discussion with content designers and more broadcasting of the change. Let's add a quick override here and remove it once the product team have switched to the new taxonomy.

## Before
Screenshot 2020-04-07 at 11 44 31

## After
Screenshot 2020-04-07 at 11 44 18

### DIFF
--- a/app/presenters/content_item_subscription_presenter.rb
+++ b/app/presenters/content_item_subscription_presenter.rb
@@ -3,6 +3,14 @@ class ContentItemSubscriptionPresenter
     @content_item = content_item
   end
 
+  def title
+    if content_item["content_id"] == "c4cd0e8a-3ae1-4385-8936-1cfafe5031fb"
+      "Coronavirus (COVID-19)"
+    else
+      content_item["title"]
+    end
+  end
+
   def description
     return "This will include: #{content_item['description']}" if is_taxon?
 
--- a/app/views/content_item_signups/confirm.html.erb
+++ b/app/views/content_item_signups/confirm.html.erb
@@ -1,4 +1,4 @@
-<% content_for :title, @content_item['title'] %>
+<% content_for :title, @subscription.title %>
 
 <% if live_content_item?(@content_item) %>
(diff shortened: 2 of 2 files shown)


vanitabarrett:
Note: I've added a Trello card to remove this (and other overrides we've added) once we've switched to the new coronavirus taxonomy. https://trello.com/c/4jT3ialW/81-tech-debt-remove-coronavirus-specific-overrides-in-email-alert-frontend

app/presenters/content_item_subscription_presenter.rb
```diff
@@ -3,6 +3,14 @@ def initialize(content_item)
     @content_item = content_item
   end
 
+  def title
+    if content_item["content_id"] == "c4cd0e8a-3ae1-4385-8936-1cfafe5031fb"
+      "Coronavirus (COVID-19)"
```
laurentqro, line 8: What is the reason for overriding it rather than changing it? We want to keep displaying the original title everywhere else, and change it here only?

app/views/content_item_signups/confirm.html.erb
```diff
@@ -14,7 +14,7 @@
     <h1 class="govuk-heading-l">Sign up to get emails</h1>
 
     <p class="govuk-body">
-      Get emails when information changes on GOV.UK about: <%= @content_item['title'] %>.
+      Get emails when information changes on GOV.UK about: <%= @subscription.title %>.
```
laurentqro, line 17: 💯

PR #749, opened 2020-04-08
Update 'manage your subscriptions' content

karlchillmaid:
This PR updates the content for the _Manage your subscriptions_ journey.

The content changes aim to:

* set user expectations
* provide useful information at the most appropriate point in the journey
* improve the consistency of content elsewhere in the email notifications service

### DIFF
--- a/config/locales/subscriber_authentication.yml
+++ b/config/locales/subscriber_authentication.yml
@@ -15,7 +15,7 @@ en:
         title: That link has expired
         description: Enter your email address again to manage your subscriptions
     request_sign_in_token:
-      heading: "Check your email inbox"
+      heading: "Check your inbox"
       confirmation: "We’ve sent an email to %{address}."
       prompt: Click the link in the email to manage your subscriptions.
       delay_warning: It can take up to 15 minutes for the email to arrive.
--- a/config/locales/subscriber_authentication.yml
+++ b/config/locales/subscriber_authentication.yml
@@ -6,10 +6,10 @@ en:
       action: We’ll send you an email with a link in to manage your subscriptions.
       email_input:
         label: Your email address
-      missing_email: "Enter your email address."
-      invalid_email: "This does not look like a valid email address – check you’ve entered it correctly."
+      missing_email: "Enter your email address"
+      invalid_email: "This does not look like a valid email address – check you’ve entered it correctly"
       email_error:
-        title: We could not sign you in
+        title: We cannot send you an email
         description: There’s a problem with your email address
       bad_token_error:
         title: That link has expired
--- a/config/locales/subscriber_authentication.yml
+++ b/config/locales/subscriber_authentication.yml
@@ -2,7 +2,8 @@ en:
   subscriber_authentication:
     heading: Manage your email subscriptions
     sign_in:
-      description: We’ll send you an email with a link in to manage your subscriptions.
+      description: For security, we need you to confirm your email address.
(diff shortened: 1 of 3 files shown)


config/locales/subscriber_authentication.yml
```diff
@@ -15,4 +15,4 @@ en:
         description: Enter your email address again to change your subscription
     request_sign_in_token:
       confirmation: "We’ve sent an email to %{address}."
-      prompt: Check your inbox and click on the link to manage your subscriptions.
+      prompt: Click the link in the email to manage your subscriptions. It can take up to 15 minutes for the email to arrive.
```
karlchillmaid, line 18: Would be better if we could split this into 2 paragraphs.

config/locales/subscriber_authentication.yml
```diff
@@ -1,18 +1,18 @@
...
     sign_in:
-      description: We’ll send you an email with a link in to manage your subscriptions.  There may be a delay of up to 15 minutes before you receive this email.
+      description: We’ll send you an email with a link in to manage your subscriptions.
       email_input:
         label: What’s your email address?
       missing_email: "Please enter your email address."
       invalid_email: "This doesn’t look like a valid email address – check you’ve entered it correctly."
       email_error:
-        title: We weren’t able to sign you in
+        title: We could not sign you in
```
karlchillmaid, line 11: Is ```sign you in``` the right concept here?
If a user does not have an account, what are they signing in to?

karlchillmaid: @vanitabarrett do you have any thoughts on this?

config/locales/subscriber_authentication.yml
```diff
@@ -1,18 +1,18 @@
 en:
   subscriber_authentication:
-    heading: Manage your subscriptions
+    heading: Manage your email subscriptions
```
karlchillmaid, line 3: @vanitabarrett Is it possible to set different headings for the 2 screens in this journey?

1. Manage your email subscriptions
2. Check your emails

vanitabarrett: @karlchillmaid Yep, that looks possible - I'll take a look!

vanitabarrett: @karlchillmaid I've added a commit with this change 👍

config/locales/subscriber_authentication.yml
vanitabarrett: @karlchillmaid I guess from a user point of view this process is similar to signing in - they're signing in to the 'manage subscriptions' area of GOV.UK. 

But it does seem like this might not be the right wording for this particular error message because at this point they're just trying to validate their email address. Even if they get past this step, they still need to click a link in an email to actually sign them in.

Screenshot 2020-04-14 at 09 53 58

Perhaps the error message could just be about the email address not being valid?

karlchillmaid: @vanitabarrett Working on error messages at the moment. We may want to put them through as a separate PR if it looks like they're going to hold this one up.

karlchillmaid: @vanitabarrett I've made a couple of small changes to the errors. This addresses the 'sign in' issue.

There's more we can do to improve these errors, but I think we'll need more input from design. We can address that later if we think it's a priority.

benthorner:
Card this relates to: https://trello.com/c/Psx11gaR/68-first-iteration-of-content-updates-for-the-manage-subscriptions-journey

PR #770, opened 2020-05-05
Show no description for taxon when it has none

laurentqro:
Currently, when taxon description is empty, the prefix "This will include:" is displayed. This makes sure we don't.

We are implementing this fix until we can find a longer-term solution for what is a somewhat flawed behaviour.

To cite @benthorner:

> Example of how this doesn't make sense: https://www.gov.uk/email-signup/?topic=/education/education-in-prisons

> Example of how showing the description at all is a bit questionable: https://www.gov.uk/email-signup?link=/government/organisations/government-digital-service

> Perhaps we just want to ditch the description entirely?

---

https://trello.com/c/V2xPGN2t/246-fix-this-will-include-when-taxon-description-is-empty

### DIFF
--- a/spec/unit/presenters/content_item_subscription_presenter_spec.rb
+++ b/spec/unit/presenters/content_item_subscription_presenter_spec.rb
@@ -8,6 +8,12 @@ RSpec.describe ContentItemSubscriptionPresenter do
         presenter = described_class.new(fake_taxon)
         expect(presenter.description).to eq "This will include: description of foo-id"
       end
+
+      it "returns nothing when there is no description" do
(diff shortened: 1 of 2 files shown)


spec/unit/presenters/content_item_subscription_presenter_spec.rb
```diff
@@ -8,6 +8,12 @@
         presenter = described_class.new(fake_taxon)
         expect(presenter.description).to eq "This will include: description of foo-id"
       end
+
+      it "shows no description" do
```
benthorner, line 12:
```suggestion
      it "returns nothing there is no description" do
```

PR #784, opened 2020-05-17
Run RuboCop as part of `bundle exec rake`

issyl0:
- There's quite a bit going on here. See individual commits for details.

https://trello.com/c/5Gob6WUL/147-for-everyone-ensure-that-rubocop-runs-as-part-of-bundle-exec-rake-as-one-of-the-default-tasks

### DIFF
--- a/lib/tasks/rspec.rake
+++ b/lib/tasks/rspec.rake
@@ -1,8 +1,8 @@
 begin
-  require 'rspec/core/rake_task'
+  require "rspec/core/rake_task"
 
   RSpec::Core::RakeTask.new(:spec)
 
-  task :default => :spec
+  task default: :spec
 rescue LoadError
 end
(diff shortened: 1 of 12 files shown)


lib/tasks/rspec.rake
```diff
@@ -1,8 +1,4 @@
-begin
-  require "rspec/core/rake_task"
-
-  RSpec::Core::RakeTask.new(:spec)
-
-  task default: :spec
-rescue LoadError
+desc "RSpec"
```
benthorner, line 1: Rather than make this different to other repos, we can get rid of it altogether if we just [put rspec-rails into the dev+test group](https://github.com/alphagov/email-alert-frontend/blob/master/Gemfile#L24).

issyl0: Thanks. Done!

PR #806, opened 2020-05-28
Support content item signup for service manual

benthorner:
https://trello.com/c/KNWv9Dgx/258-not-ready-to-start-deprecate-and-remove-legacy-email-signup-pages

Previously these relied on the 'email_alert_signup' functionality in this
app, which we'd like to remove, since most formats now use the generic
'content_item_signup' functionality. This adds support for creating
subscriptions based on the content item.

Please see the commits for more details.

### DIFF
--- a/app/controllers/content_item_signups_controller.rb
+++ b/app/controllers/content_item_signups_controller.rb
@@ -35,7 +35,14 @@ class ContentItemSignupsController < ApplicationController
 
 private
 
-  PERMITTED_CONTENT_ITEMS = %w[taxon organisation ministerial_role person topical_event topic].freeze
+  PERMITTED_CONTENT_ITEMS = %w[taxon
+                               organisation
+                               ministerial_role
+                               person
+                               topic
+                               topical_event
+                               service_manual_topic
+                               service_manual_service_standard].freeze
 
   def require_content_item_param
     unless valid_content_item_param?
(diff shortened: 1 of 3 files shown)


app/controllers/content_item_signups_controller.rb
```diff
@@ -35,7 +35,13 @@ def create
...
 private
 
-  PERMITTED_CONTENT_ITEMS = %w[taxon organisation ministerial_role person topical_event].freeze
+  PERMITTED_CONTENT_ITEMS = %w[taxon
+                               organisation
+                               ministerial_role
+                               person
+                               topical_event
+                               service_manual_topic
+                               service_manual_service_standard].freeze
```
vanitabarrett, line 44: This needs to be rebased following Kelvin's [change to support topics](https://github.com/alphagov/email-alert-frontend/pull/789)

app/controllers/content_item_signups_controller.rb
```diff
@@ -26,7 +26,7 @@ def confirm
   def create
     signup = ContentItemSubscriberList.new(content_item.to_h)
```
vanitabarrett, line 27: Could we move this within the if statement too, so we're only doing this once we know it's present?

benthorner: Oh yes, good point - updated.

app/controllers/content_item_signups_controller.rb
benthorner: Sure - I've added another commit after doing that. Can you take another look?

PR #808, opened 2020-05-29
Show error when user tries to confirm email-signup without a topic

vanitabarrett:
Previously, we redirected to root (GOVUK homepage) for all instances when requests to new/confirm/create had invalid or missing params.
There are cases where we may still want to do this, so I've tried to keep those as-is.

However, if a user tries to click through from the initial taxon sign-up page without selecting a topic, we now show a useful error message and redirect them back to the selection page.

Possible scenarios:
- New (without topic param): redirect to /
- New (with invalid topic param): 404
- __Confirm (without topic param, submitted via form with current_topic): show error__ 
- Confirm (without topic param, direct link so no current_topic): redirect to /
- Confirm (with invalid topic param): 404

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -8,7 +8,7 @@ class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
 
   rescue_from ActionController::InvalidAuthenticityToken, with: :invalid_token
-  rescue_from GdsApi::HTTPNotFound, with: :error_not_found
+  #rescue_from GdsApi::HTTPNotFound, with: :error_not_found
   rescue_from GdsApi::HTTPForbidden, with: :forbidden
   rescue_from GdsApi::HTTPGone, with: :gone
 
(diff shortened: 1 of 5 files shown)


benthorner:
Just a couple of thoughts that might make this easier:

- The current redirect behaviour is not worth preserving, so we could just change that.
- We shouldn't return a 404, because the `/email-signup` path does exist; we could return a 400
- Could we pre-select the taxon? The user has already made a selection on the previous page

It looks like this wasn't really thought about [when we removed the original pre-selection](https://github.com/alphagov/email-alert-frontend/commit/ba47b1e0cc6db5105fd7646212b951e2e7c54c1d#diff-f03d1ffe3e03d626fa1b05506e99ec17).

vanitabarrett:
Closing this PR, it is replaced by these 2 PRs instead:

https://github.com/alphagov/email-alert-frontend/pull/811
https://github.com/alphagov/email-alert-frontend/pull/812

PR #812, opened 2020-06-03
Remove homepage redirect

vanitabarrett:
Related to https://trello.com/c/XxMJ77lG/302-add-default-to-taxon-selection-page and https://github.com/alphagov/email-alert-frontend/pull/811

## What
Removes the code which redirects a user to the GOV.UK homepage if something goes wrong with email signup and instead raises a 400 Bad Request error.

Summary:

| Scenario | Previous Behaviour | New Behaviour | Example |
| --- | --- | --- | --- |
| New (invalid topic) | redirect to / | returns 400 | /email-signup/?topic=123 |
| New (missing topic) | redirect to / | returns 400 | /email-signup |
| New (not a taxon) | redirect to / | returns 400 | /email-signup/?topic=/cma-cases |
| Confirm (invalid topic) | redirect to / | returns 400 | /email-signup/confirm?utf8=%E2%9C%93&topic=123 |
| Confirm (missing topic) | redirect to / | returns 400 | /email-signup/confirm?utf8=%E2%9C%93 |

## Why
If something goes wrong, throwing a user back to the GOV.UK homepage doesn't seem like a useful thing to do. The existing code returns a 404.

### DIFF
--- a/app/controllers/content_item_signups_controller.rb
+++ b/app/controllers/content_item_signups_controller.rb
@@ -45,8 +45,7 @@ private
 
   def require_content_item_param
     unless valid_content_item_param?
-      redirect_to "/"
-      false
+      bad_request
     end
   end
 
@@ -83,8 +82,7 @@ private
 
   def validate_document_type
     unless PERMITTED_CONTENT_ITEMS.include?(content_item["document_type"])
-      redirect_to "/"
-      false
+      bad_request
     end
   end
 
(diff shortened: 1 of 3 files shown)


app/controllers/content_item_signups_controller.rb
```diff
@@ -83,8 +82,7 @@ def handle_redirects
 
   def validate_document_type
```
benthorner, line 83: Minor: we could simplify the number of `before_actions` and combine these into a single `validate_signup_params` action, or something.

PR #823, opened 2020-06-16
Add extreme EmailVolume value

chao-xian:
The only taxon we are getting extreme volumes for are notifications
for content tagged with the Coronavirus taxon. We want to inform
the user what these numbers are, so that they can get decide whether
they still want to get immediate notifications. The min value is
based on a daily average, and the maximum is on a day's maximum
content changes during the first couple of weeks of the taxon's
existence.

https://trello.com/c/rcC3U4Ah

### Before
Screenshot 2020-06-16 at 5 27 34 pm

### After
Screenshot 2020-06-16 at 5 27 17 pm

### DIFF
--- a/app/models/email_volume/taxon_weekly_email_volume.rb
+++ b/app/models/email_volume/taxon_weekly_email_volume.rb
@@ -1,5 +1,6 @@
 module EmailVolume
   class TaxonWeeklyEmailVolume
+    EXTREME = "200 to 800".freeze
     HIGH = "40 to 60".freeze
     MEDIUM = "0 to 20".freeze
     LOW = "0 to 5".freeze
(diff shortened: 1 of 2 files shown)


vanitabarrett:
Can we update the tests? https://github.com/alphagov/email-alert-frontend/blob/ceec1b56115e8bc8952eb0e249600a8e5f8caf25/spec/models/weekly_email_volume_spec.rb

chao-xian:
@vanitabarrett [done](https://github.com/alphagov/email-alert-frontend/commit/3dcda0e2dfec5d0c19378b2ac0c19487c40a1272) Don't know how I missed that. I was originally puzzled as to why `EmailVolume::TaxonWeeklyEmailVolume` didn't have a test and was wondering where would be best to add one. Didn't realise it was covered in one for `EmailVolume::WeeklyEmailVolume`.

PR #833, opened 2020-06-26
Harden rate limits for verification emails

benthorner:
https://trello.com/c/fCrWTqFj/372-incident-action-rate-limit-email-authentication-emails-by-address

### DIFF
--- a/app/services/verify_subscriber_email_service.rb
+++ b/app/services/verify_subscriber_email_service.rb
@@ -3,8 +3,14 @@ class VerifySubscriberEmailService < ApplicationService
 
   class RatelimitExceededError < StandardError; end
 
-  MINUTELY_THRESHOLD = 3 # TODO: update this based on data and comment
-  HOURLY_THRESHOLD = 20 # TODO: update this based on data and comment
+  # This allows for up to 2 retries, to account for users
+  # not getting the email quickly enough and resubmitting.
+  MINUTELY_THRESHOLD = 4
+
+  # This allows for up to 10 logins to manage subscriptions
+  # per hour, to account # for the user needing to return
+  # to manage their # subscriptions on different devices.
+  HOURLY_THRESHOLD = 11
 
   def initialize(address)
     @address = address
--- a/app/services/verify_subscription_email_service.rb
+++ b/app/services/verify_subscription_email_service.rb
@@ -1,8 +1,15 @@
 class VerifySubscriptionEmailService < ApplicationService
   class RatelimitExceededError < StandardError; end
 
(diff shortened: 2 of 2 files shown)


app/services/verify_subscriber_email_service.rb
```diff
@@ -3,8 +3,14 @@ class VerifySubscriberEmailService < ApplicationService
...
-  MINUTELY_THRESHOLD = 3 # TODO: update this based on data and comment
-  HOURLY_THRESHOLD = 20 # TODO: update this based on data and comment
+  # This allows for up to 2 retries, to account for users
+  # not getting the email quickly enough and resubmitting.
+  MINUTELY_THRESHOLD = 4
+
+  # This allows for up to 5 retries per hour, to account
+  # for the user needing to return to manage their
+  # subscriptions on different devices in the same hour.
+  HOURLY_THRESHOLD = 6
```
kevindew, line 13: This feels a bit low, particularly since this is the less vulnerable end-point for spamming. Given an hour is quite a long time it feels realistic you could do a fair more than this. Does doubling it (or matching the other one at 11) seem reasonable?

app/services/verify_subscription_email_service.rb
```diff
@@ -1,8 +1,15 @@
...
-  HOURLY_THRESHOLD = 20 # TODO: update this based on data and comment
+  # This allows for up to 2 retries, to account for users
+  # not getting the email quickly enough and resubmitting,
+  # and for users who subscribe to several things quickly.
+  MINUTELY_THRESHOLD = 4
+
+  # This allows for up to 10 subscriptions per hour. Analysis
+  # shows 99% of subscribers have 10 or fewer active subs,
+  # so this covers someone creating all their subs in bulk.
+  HOURLY_THRESHOLD = 11
```
kevindew, line 12: Nice digging, it'd be be good to have the query used in the commit message.

benthorner: Sure, added.

app/services/verify_subscriber_email_service.rb
benthorner: For the lack of specific data, yes. It now matches the other journey.

benthorner:
@kevindew this should be ready for another look.

kevindew:
Thanks @benthorner, I can see over the weekend we did serve quite a lot of 429 (excluding 2 particular IP addresses there were 40 served total, to 29 different IP addresses). Can we have a look at the request patterns there to determine if we need to tweak these rules further?

benthorner:
Sure, I've added a note about that to the card 👍

PR #834, opened 2020-06-26
Track selected choices across email signup journey

laurentqro:
Adds analytics tracking to email sign-up journey.

https://trello.com/c/mTcA2hFA/307-add-tracking-to-email-sign-up-journey
https://docs.google.com/presentation/d/1sPjnEiA9w7Ctrl-7HLkOj9WNwJ_6ShnP3j7g06JS7n0/edit#slide=id.p


app/assets/javascripts/email-analytics.js
```diff
@@ -0,0 +1,17 @@
...
+  'use strict'
+
+  window.GOVUK = window.GOVUK || {}
+
+  function canTrack () {
+    return !!GOVUK.analytics
+  }
+
+  // The EmailAnalytics module is a wrapper around GOVUK.analytics
+  GOVUK.EmailAnalytics = {
```
vanitabarrett, line 11: I'm not sure if this is needed? Can we just use `GOVUK.analytics.trackEvent` directly? I'm not sure if the wrapper is adding much.

spec/javascripts/modules/track-email-alert-signup.spec.js
```diff
@@ -0,0 +1,48 @@
...
+
+    tracker = new GOVUK.Modules.TrackEmailAlertSignupChoices();
+    tracker.start($element)
+  })
+
+  afterEach(function () {
+    GOVUK.EmailAnalytics.trackEvent.calls.reset()
+  })
+
+  it('tracks selected radio buttons when clicking submit', function () {
```
vanitabarrett: Should we test the scenario when no option has been selected?

app/views/content_item_signups/confirm.html.erb
```diff
@@ -31,9 +31,18 @@
...
-    <%= form_tag(:action => :create) do %>
+    <%= form_tag(action: :create) do %>
       <%= hidden_field_tag 'link', @content_item['base_path'] %>
-      <%= submit_tag('Sign up', class: 'govuk-button') %>
+      <%= render 'govuk_publishing_components/components/button', {
+        text: 'Sign up',
+        margin_bottom: true,
+        data_attributes: {
+          module: 'auto-track-event',
+          track_category: 'email_subscriptions',
```
vanitabarrett, line 41: For these tracking attributes to work, you can just use module: "track-click" instead of "auto-track-event" as we're doing standard tracking here (nothing custom).

app/views/subscriptions/new_address.html.erb
```diff
@@ -48,7 +48,7 @@
         data_attributes: {
           module: 'auto-track-event',
```
vanitabarrett, line 49: This should be "track-click"

app/views/subscriptions/new_address.html.erb
```diff
@@ -29,7 +29,7 @@
       <%= t("subscriptions.new_address.summary.#{@frequency}") %>
     </p>
 
-    <%= form_tag verify_subscription_path, method: :post, class: "checklist-email-signup", novalidate: "novalidate" do %>
```
vanitabarrett, line 32: I think we need the `novalidate: "novalidate"` so we dont show the default browser error if someone tries to submit the form without an email address?

app/assets/javascripts/modules/track-email-alert-signup-choices.js
```diff
@@ -0,0 +1,42 @@
...
+              ? checkedOptionLabel
+              : $checkedOption.val()
+
+            options = { transport: 'beacon', label: eventLabel }
+
+            GOVUK.EmailAnalytics.trackEvent(action, category, options)
+          })
+        } else {
+          options = { transport: 'beacon', label: 'no choice' }
+          GOVUK.EmailAnalytics.trackEvent(action, category, options)
```
vanitabarrett, line 37: Can you double check these are being passed in the right order? When I tested on integration, the category and action were the wrong way around.

app/assets/javascripts/modules/track-email-alert-signup-choices.js
```diff
@@ -0,0 +1,43 @@
...
+            $checkedOption = $(this)
+            var checkedOptionId = $checkedOption.attr('id')
+            var checkedOptionLabel = $submittedForm.find('label[for="' + checkedOptionId + '"]').text().trim()
+            eventLabel = checkedOptionLabel.length
+              ? checkedOptionLabel
+              : $checkedOption.val()
+
+            options = { transport: 'beacon', label: eventLabel }
+
+            GOVUK.analytics.trackEvent(category, action, options)
```
vanitabarrett, line 33: Now we're not using the wrapper, I think we need to check that GOVUK.analytics is available before we call it. (same comment for a couple of lines below too)

app/assets/javascripts/modules/track-email-alert-signup-choices.js
```diff
@@ -0,0 +1,43 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (global, GOVUK) {
+  'use strict'
+
+  var $ = global.jQuery
```
kevindew, line 7: There's a movement, that is gathering steam, to not write new code that requires jQuery based on the [gds-way](https://gds-way.cloudapps.digital/manuals/programming-languages/js.html#jquery). 

Since these changes introduce jQuery to this project it'd probably save some pain if this was in vanilla JS. I know this creates a bit of a pain as it's a port though it looks like jQuery is only used for DOM traversal and setting an event handler.

app/assets/javascripts/modules/track-email-alert-signup-choices.js
```diff
@@ -0,0 +1,43 @@
...
+    function track (element) {
+      element.on('submit', function (event) {
+        var $checkedOption, eventLabel, options
+        var $submittedForm = $(event.target)
+        var $checkedOptions = $submittedForm.find('input:checked')
+        var category = $submittedForm.data('track-category')
+        var action = $submittedForm.data('track-action')
+
+        if ($checkedOptions.length) {
+          $checkedOptions.each(function (index) {
```
kevindew, line 23: Are you expecting to need a loop here? I thought the things being were tracked are radio buttons.

config/initializers/jasmine_rails.rb
```diff
@@ -0,0 +1,11 @@
+if defined?(JasmineRails)
```
kevindew, line 1: Do we need this file / condition?

It seems a bit strange that we apply this only when we have JasmineRails defined when it's available in both dev and test meaning this would, if I'm understanding correct, apply to every single request in dev and test?

app/assets/javascripts/application.js
```diff
@@ -1,6 +1,11 @@
+//= require govuk_publishing_components/dependencies
```
kevindew, line 1: This file shouldn't be included for an app that has slimmer (such as this) as it introduces the dependencies that static already provides (we'll end up bundling jQuery twice on GOV.UK for instance)

To get these dependencies in a test environment only you can take an approach like: https://github.com/alphagov/smart-answers/commit/bda1c918f6d179925f91766bce6fd8ce27a7063d

app/views/content_item_signups/confirm.html.erb
```diff
@@ -31,9 +31,18 @@
       You can choose how often you want to receive emails.
     </p>
 
-    <%= form_tag(:action => :create) do %>
+    <%= form_tag(action: :create) do %>
```
kevindew, line 34: Nice 👍

app/assets/javascripts/email-analytics.js
```diff
@@ -0,0 +1,17 @@
...
+    trackEvent: function trackEvent () {
+      if (!canTrack() || !GOVUK.analytics.trackEvent) { return }
```
kevindew, line 13: You could probably simplify this code a bit by removing the need for extra function and changing this to:

```suggestion
      if (!GOVUK.analytics || !GOVUK.analytics.trackEvent) { return }
```

spec/javascripts/support/jasmine.yml
```diff
@@ -0,0 +1,30 @@
+# path to parent directory of src_files
+# relative path from Rails.root
+# defaults to app/assets/javascripts
+src_dir: "app/assets/javascripts"
+
+# list of file expressions to include as source files
+# relative path from src_dir
+src_files:
+  - "../../../spec/javascripts/helpers/jquery-3.5.1.js"
```
kevindew, line 9: Where does this come from? I don't think it's needed and it's a very different version of jQuery to what GOV.UK uses 1.12

app/assets/javascripts/modules/track-email-alert-signup-choices.js
kevindew: /cc @alex-ju it'd be useful to get your perspective on what approach we should be taking for porting jQuery code into projects which currently don't have jQuery. I do feel a bit mean about making the suggestion of changing but does seem to be reflective of the govuk-frontenders stance as per yesterdays meeting?

alex-ju: @laurentqro we recently made a decision at the community level that we must stop introducing more jQuery code in our efforts to move away from jQuery and align with the GDS way as Kevin mentioned above. Since this script is relatively straight-forward, I will have to ask for it to be written in pure JavaScript and use the [reference module pattern](https://github.com/alphagov/govuk_publishing_components/blob/master/docs/javascript-modules.md#module-structure). [Here](https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/javascripts/govuk_publishing_components/components/contextual-guidance.js) is a decent example of how the pattern is used in practice and I am happy to answer any questions related to this refactor.

laurentqro:
@kevindew @alex-ju I opened a new branch with an easy-to-follow commit history for your reviewing pleasure: https://github.com/alphagov/email-alert-frontend/pull/838.

@kevindew I was somehow unable to load the test dependencies in isolation. Your suggested solution suggested loading them conditionally in the application layout view, but my tests are unit tests. Suggestions welcome.

PR #838, opened 2020-07-07
Track email signup user journey

laurentqro:
Adds analytics tracking to email sign-up journey.

https://trello.com/c/mTcA2hFA/307-add-tracking-to-email-sign-up-journey

https://docs.google.com/presentation/d/1sPjnEiA9w7Ctrl-7HLkOj9WNwJ_6ShnP3j7g06JS7n0/edit?ts=5efc8fc7#slide=id.p

Uses the GDS way module structure, as per https://gds-way.cloudapps.digital/manuals/programming-languages/js.html#modules.


package-lock.json
```diff
@@ -0,0 +1,878 @@
+{
```
alex-ju, line 1: Really appreciate setting up Standard for listing 🙇  We're trying to use yarn in Rails projects – as I understand it comes out of the box. There's a good reference for the [package.json in content-publisher](https://github.com/alphagov/content-publisher/blob/master/package.json).

Gemfile.lock
```diff
@@ -121,6 +121,15 @@ GEM
       domain_name (~> 0.5)
     i18n (1.8.3)
       concurrent-ruby (~> 1.0)
+    jasmine (3.5.1)
+      jasmine-core (~> 3.5.0)
+      phantomjs
```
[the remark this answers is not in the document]
kevindew: Jasmine bundles phantomjs unfortunately. We can't avoid it being a dependency 😞  we [end up with it in Content Publisher too](https://github.com/alphagov/content-publisher/blob/d41fabb5ce00f137ed6b409c130c51c499f91b9f/Gemfile.lock#L193). Might not be something resolved until we reach the distant land of webpacker and jest/karma.

app/assets/javascripts/application.js
```diff
@@ -1,6 +1,9 @@
+//= require govuk_publishing_components/dependencies
```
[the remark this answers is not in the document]
kevindew: I recall this being needed for the unit tests on a comment on the previous PR.

The approach you can take to include this in only tests is to add it to the start of jasmine.yml src definition. For example:

```yml
src_files:
  - assets/govuk_publishing_components/dependencies.js
  - assets/application.js
```

and then would also need referencing in config/initializers/assets.rb to flag it as a file that needs precompiling.

Gemfile
```diff
@@ -16,6 +16,8 @@ gem "uglifier", "~> 4.2"
 
 group :development, :test do
   gem "govuk_test", "~> 1.0"
+  gem "jasmine", "~> 3.5.1"
+  gem "jasmine_selenium_runner", "~> 3", require: false
```
kevindew, line 20: Sorry to be a pain again, we pushed up a new approach to pulling in jasmine_selenium_runner that avoids the need for `require: false` that we're going to try popularise before the `require: false` technique spreads too far (as it essentially fights the gems configuration). Details are here: https://github.com/alphagov/content-publisher/pull/2110

app/assets/javascripts/modules/track-email-alert-signup-choices.js
```diff
@@ -0,0 +1,46 @@
...
+    }
+  }
+
+  TrackEmailAlertSignupChoices.prototype.forwardToAnalytics = function () {
+    if (!GOVUK.analytics || !GOVUK.analytics.trackEvent) { return }
+    return GOVUK.analytics.trackEvent.apply(GOVUK, arguments)
+  }
+
+  Modules.TrackEmailAlertSignupChoices = TrackEmailAlertSignupChoices
+})(window.GOVUK.Modules)
```
kevindew, line 46: This looks great 👍

package.json
```diff
@@ -0,0 +1,28 @@
...
+  "directories": {
+    "doc": "docs",
+    "lib": "lib"
+  },
+  "dependencies": {},
+  "devDependencies": {
+    "eslint": "^7.3.1"
+  },
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
```
kevindew, line 15: Probably worth removing this line.

package.json
```diff
@@ -0,0 +1,28 @@
+{
```
kevindew, line 1: We should mark this with private: true to indicate it's not a public package.

package.json
```diff
@@ -0,0 +1,28 @@
...
+  "name": "email-alert-frontend",
+  "version": "1.0.0",
+  "description": "A frontend for creating and managing email subscriptions.",
```
kevindew, line 4: Somewhat moot given this is private but should reference GOV.UK in the description.

package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "email-alert-frontend",
+  "version": "1.0.0",
```
kevindew, line 3: I don't think we want this since we won't maintain it (I believe it's still valid without a version)

package.json
```diff
@@ -0,0 +1,28 @@
...
+  "description": "A frontend for creating and managing email subscriptions.",
+  "main": "index.js",
+  "directories": {
+    "doc": "docs",
+    "lib": "lib"
+  },
```
kevindew, line 9: I don't think we want/need the main and directories references here.

package.json
```diff
@@ -0,0 +1,28 @@
...
+  "dependencies": {},
+  "devDependencies": {
+    "eslint": "^7.3.1"
```
kevindew, line 12: Should this actually be a dependency on [standard](https://www.npmjs.com/package/standard)?

package.json
```diff
@@ -0,0 +1,28 @@
...
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "lint": "standard app/assets/javascripts/{*.js,**/*.js} spec/javascripts/{*.js,**/*.js}"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/alphagov/email-alert-frontend.git"
+  },
+  "author": "GDS",
+  "license": "ISC",
```
kevindew, line 23:
```suggestion
  "license": "MIT",
```

as per: https://github.com/alphagov/email-alert-frontend/blob/master/LICENCE

package.json
```diff
@@ -0,0 +1,28 @@
...
+  },
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "lint": "standard app/assets/javascripts/{*.js,**/*.js} spec/javascripts/{*.js,**/*.js}"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/alphagov/email-alert-frontend.git"
+  },
+  "author": "GDS",
```
kevindew, line 22:
```suggestion
  "author": "Government Digital Service",
```

package.json
```diff
@@ -0,0 +1,28 @@
...
+  "license": "ISC",
+  "bugs": {
+    "url": "https://github.com/alphagov/email-alert-frontend/issues"
+  },
+  "homepage": "https://github.com/alphagov/email-alert-frontend#readme"
```
kevindew, line 27: I don't think we want/need these bugs and homepage links

Rakefile
```diff
@@ -6,4 +6,4 @@ require File.expand_path("config/application", __dir__)
 Rails.application.load_tasks
 
 Rake::Task[:default].clear unless Rails.env.production?
-task default: %i[lint spec]
+task default: %i[lint spec jasmine:ci]
```
kevindew, line 9: We should also expand the lint task to include the JS linting, example: https://github.com/alphagov/content-publisher/blob/d41fabb5ce00f137ed6b409c130c51c499f91b9f/lib/tasks/lint.rake#L6

Gemfile.lock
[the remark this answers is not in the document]
alex-ju: Ah, right. Sorry about the confusion then.

app/assets/javascripts/application.js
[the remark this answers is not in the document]
laurentqro: This works, thanks @kevindew ! :tada:

laurentqro:
> Great, looks good to me. Might be worth confirming with a PA that it works like they expect in integration before merging.

Tested on integration and approved by Andrew. 👌

PR #849, opened 2020-07-27
Make app consistent with updates

cdccollins:
As part of the work upgrading the GDS Rails apps thought it would be best to make all apps consistent.


EGiataganas:
Hey @cdccollins this looks great, I've compared the PR changes with [railsdiff][1] and there are some changes that didn't happen with `bundle exec rails app:update` that could also be done:

1. Remove `bin/bundle`, `bin/update`
2. In the `config/application.rb` remove the unused (commented) Railties require statements https://github.com/alphagov/travel-advice-publisher/pull/919/commits/c59db4c22030b33b458e0b88e91c9f807b7ec123
3. Check whether we need to explicitly add the local dev link http://email-alert-frontend.dev.gov.uk/ to the config hosts in the `config/environments/development.rb` https://github.com/alphagov/travel-advice-publisher/pull/919/commits/2689aa4
4. In the `config/environments/test.rb` the `config.cache_classes = true` has changed to `false` and there is `config.action_view.cache_template_loading = true`
5. Remove `config/initializers/new_framework_defaults_5_1.rb` & `config/initializers/new_framework_defaults_6_0.rb` as per the reason in https://github.com/alphagov/travel-advice-publisher/commit/8b0745e
6. This app includes `config/cable.yml`  but it doesn't use it. So I think it's better to delete it.
7. In the `config/puma.rb` Rails 6 introduce `has max_threads_count` and `min_threads_count`. It will be good to update it
8. In the `config/spring.rb` the `%w[]` needs to be updated to `Spring.watch()`
```ruby
Spring.watch(
  ".ruby-version",
  ".rbenv-vars",
  "tmp/restart.txt",
  "tmp/caching-dev.txt"
)
```
9. Update `app/views/layouts/application.html.erb`to include `csp_meta_tag` after `csrf_meta_tags`.

[1]: http://railsdiff.org/5.2.4.3/6.0.3.2

EGiataganas (approved):

PR #858, opened 2020-08-20
Verify user owns subscription before updating

kevindew:
This fixes a problem where using the endpoint to change_frequency was
raising an error if a subscription was modified that didn't belong to a
particular subscriber.


spec/controllers/subscriptions_management_controller_spec.rb
```diff
@@ -101,27 +101,31 @@
...
 
-      it "returns 404" do
-        get :update_frequency, params: { id: non_existant_subscription_id }, session: session_data
-        expect(response).to have_http_status(:not_found)
-      end
+      get :update_frequency, params: { id: subscription_id }, session: session_data
+      expect(response).to have_http_status(:not_found)
+    end
+
+    it "returns 404 when the requested subscription doesn't exist" do
```
benthorner, line 125: We seen to be missing a `404` test for the POST endpoint.

PR #866, opened 2020-09-08
Move radio button captions into legends for those radio button fieldsets

owenatgov:
## What
Moves captions for radio button collections into the the legend for that radio button fieldset.

## Why
Currently these captions exist outside the fieldset legends as styled p tags. This makes the caption for the radio fieldset not programatically determinable, failing WCAG SC 4.1.2. By moving it into the legend associated with the radio fieldset it clearly associates that caption with the radio buttons and allows us to make more efficient use of the components gem radio buttons API.

### Before
Screenshot 2020-09-07 at 10 46 50

### After
Screenshot 2020-09-07 at 10 46 37

**Card:** https://trello.com/c/Np871Jwm/335-email-alerts-missing-legend

### DIFF
--- a/app/views/content_item_signups/new.html.erb
+++ b/app/views/content_item_signups/new.html.erb
@@ -22,9 +22,10 @@
                    track_category: "email_subscriptions",
                  }) do %>
 
-      <p class="app-email-option">Everything published to GOV.UK about</p>
+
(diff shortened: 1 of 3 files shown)


injms (approved):
📨  👍

PR #883, opened 2020-10-02
Remove redundant government navigation

benthorner:
https://trello.com/c/233ZtYfR/504-design-how-to-discourage-new-subscriptions-for-individual-emails

Previously we used to show a bunch of links on the initial page for
a '/government/*/email-signup' type subscription. Since only travel
advice still uses the '*/email-signup' workflow [1], we can safely
delete this untested and unused feature.

Removing this now will simplify our upcoming work around revamping
the signup pages; it's an unnecessary complication.

[1]: https://github.com/alphagov/email-alert-frontend#signup

### DIFF
--- a/app/models/email_alert_signup.rb
+++ b/app/models/email_alert_signup.rb
@@ -26,14 +26,6 @@ class EmailAlertSignup
       .find_or_create_subscriber_list(subscription_params)
   end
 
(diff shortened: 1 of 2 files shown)


kevindew (approved):
Nice catch

PR #905, opened 2020-11-04
Reorder and revise frequency options

benthorner:
https://trello.com/c/CJnbG5rY/582-update-frequency-selection-radio-buttons-experiment

This changes the order and wording of the frequency options to try
and discourage users from selecting "immediately".

Note that this change applies equally to selecting the frequency
for new and existing subscriptions.

Note that this removes the default selected frequency for new
Coronavirus subscriptions, since this is a design no-no, and should
be superseded by the overall change here.

Note that this removes the test for a helper, which was far too
coupled to an exact output, and would only now serve to test a
very simple (and optional) behaviour where any previously selected
frequency was replayed to the user.

## Before

Screenshot 2020-11-04 at 15 36 55
Screenshot 2020-11-04 at 15 37 33

## After

Screenshot 2020-11-04 at 15 39 37
Screenshot 2020-11-04 at 15 39 55


1pretz1 (approved):
🎖️

PR #912, opened 2020-11-17
Add missing h1

chris-gds:
## What? 

Updating page so it includes a `h1`, also updating `README` so startup is clearer

## Why?
[Ticket](https://trello.com/c/12elBDh4)
[There is no h1 on this page](https://www.gov.uk/email/subscriptions/new?topic_id=uk-help-and-services-in-switzerland-24e8ef8275). The main content starts with an h2.
Not a strict WCAG fail, screen reader users might find it more difficult to get to the main content and understand the structure of the page. 

## Visual Changes

Before:

Screenshot 2020-11-17 at 14 25 57

After:

Screenshot 2020-11-19 at 14 21 13

## Anything else?
Can view via this local URL

`http://email-alert-frontend.dev.gov.uk/email/subscriptions/new?topic_id=uk-help-and-services-in-switzerland`

The README has also been updated to include some clearer startup guidance (I ran into some problems: the docker `--live` flag is essential for this repo, this wasn't clear to me as I ran into errors spinning this up in isolation.

### DIFF
--- a/app/views/subscriptions/new_frequency.html.erb
+++ b/app/views/subscriptions/new_frequency.html.erb
@@ -33,6 +33,8 @@
         name: "frequency",
         id: "email-frequency-input",
         heading: t("subscriptions.new_frequency.question"),
+        is_page_heading: true,
+        heading_size: "l",
         error_message: flash[:error],
         items: frequencies(frequency_options(@topic_id)),
       } %>


README.md
```diff
@@ -4,6 +4,20 @@ A frontend for creating and managing email subscriptions.
...
+$ ./startup.sh
+```
+
+If you encounter 404 pages, you can also use [govuk-docker](https://github.com/alphagov/govuk-docker) adding the `--live` flag to use live data 
+
+```sh
+$ govuk-docker up email-alert-frontend-app-live
+```
+
+If you are using the GDS development virtual machine then the application will be available on the host at http://email-alert-frontend.dev.gov.uk/
```
kevindew, line 19: Thanks for suggesting this.

This guidance seems very specific towards running this app in isolation of the rest of the GOV.UK stack - which might be the common case for a frontend developer but not necessarily what a backend developer wants.

This guidance was recently removed in https://github.com/alphagov/email-alert-frontend/commit/d6de99f6b1104f8e02a088f394ee41cf287c66f1#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5 - with this approach applied across the GOV.UK stack. 

So rather than a small change this is probably part of a wider conversation about what should be in GOV.UK applications READMEs.

chris-gds: > This guidance seems very specific towards running this app in isolation of the rest of the GOV.UK stack 

I'm wondering if there's a better way to phrase this?  In this particular context the app _was_ run in isolation.  

I'm thinking there might be a concern about noise within this doc? Wondering if there's a halfway to reduce the length but still include the info that would have been helpful.

> which might be the common case for a frontend developer but not necessarily what a backend developer wants.

Is this app generally speaking, only used backend developers? Is more documentation problematic in some way? 
>So rather than a small change this is probably part of a wider conversation about what should be in GOV.UK applications READMEs.

Maybe yes, it might be nice to have more of a generic README Template that's consistent across all apps that includes startup in isolation.

danacotoran:
@chris-gds I notice a discrepancy between the existing `h1` styles on the email signup pages vs the new **"How often do you want to receive emails?"**. Namely that the latter seems to have significantly larger font size.

### Other email signup pages `h1`
Screenshot 2020-11-19 at 13 28 09

### New `h1` on the frequency page
image


Should the `h1` styles be consistent across the email signup journey?

chris-gds:
@danacotoran 
> Should the `h1` styles be consistent across the email signup journey?

It's a good point (will update)

`<h1 class="govuk-heading-xl">govuk-heading-xl</h1>`
`<h1 class="govuk-heading-l">govuk-heading-l</h1>`

Looking at the Design system it seems this updated code has the default `h1` styles with the surrounding journey having a smaller `h1`.

It think it harps back to other question yesterday - shouldn't all headers have the same style and size everywhere? However as pointed out the Design System  says _"change it if your pages feel unbalanced – semantic and visual hierarchy do not always need to be the same"_ - so no.

I wonder the value of having all different sized h1s

README.md
kevindew: >  Is more documentation problematic in some way?

More isn't problematic, but misleading is. One of the challenges we've had (and why the existing run guides were deleted) was that they suggested people do things in a way that wasn't consistent with the govuk-docker recommended approach, which means someone would follow the documentation and it wouldn't work.

Noise isn't so much a concern it's more about repeating things that could age badly - for example listing the govuk-docker commands in this README couples this documentation to any changes in govuk-docker.

I'm guessing what we'd want to say to tell someone how to run this application is:

```
### Running this application

This application is designed to be run within the [GOV.UK Docker](https://github.com/alphagov/govuk-docker) development environment.
```

The `startup.sh` scripts are caught as a bit of an anomaly since the retirement of the development VM. Removing them would definitely cause people problems so that isn't currently considered but it doesn't seem great to be suggesting two different approaches to running an app and making a developer choose.

It's probably worth breaking this change into a separate PR so it doesn't block the heading change.

chris-gds: >This application is designed to be run within the [GOV.UK Docker](https://github.com/alphagov/govuk-docker) development environment.

This works, can we add anything within this copy to mentioned the `--live` flag?  As this was the tiny bit of crucial info that (I don't think..) is required for all apps?

Perhaps this can be updated on [this docker page](https://github.com/alphagov/govuk-docker/blob/master/docs/compatibility.md)? I'm trying to think of a way - coming into this without touching anything else for the first time - this can be communicated

danacotoran (approved):
Looks good to me now

README.md
```diff
@@ -4,6 +4,10 @@ A frontend for creating and managing email subscriptions.
 
 ## Overview
 
+### Running the application
+
+This application is designed to be run within the [GOV.UK Docker](https://github.com/alphagov/govuk-docker) development environment, adding the `--live` flag
```
kevindew, line 9: It looks like we have an unfinished sentence here? 

The `--live` flag doesn't have a bearing on GOV.UK Docker by the way - just the startup.sh script. 

I do still think this should be a separate PR or an issue as it'd be good to get some consensus as any decisions should be things that can be applying across the suite of apps consistently.

chris-gds: Maybe a full stop?  Or additional "by'?

> This application is designed to be run within the GOV.UK Docker development environment adding the `--live` flag.
 
I'm just a bit concerned it might get lost and then someone else faces the same issue esp as we have startup documentation information across other FE apps - this feels like an anomaly.

I can raise as an issue? This at least documents it.

PR #916, opened 2020-11-23
Update common pages for new subscriptions

benthorner:

This updates the pages that are common to all new subscription flows
with a new design. Please see the commits for more details.

One general theme is that I've used 'target="_blank"' for all links.
While the design didn't specify either way, most of the links seem to
be providing additional information, where it would not make sense for
the user to navigate away.

## Before

Screenshot 2020-11-23 at 13 18 45
Screenshot 2020-11-23 at 13 18 54
Screenshot 2020-11-23 at 13 19 00

## After

Screenshot 2020-11-23 at 17 22 21

Screenshot 2020-11-23 at 13 18 20
Screenshot 2020-11-23 at 13 18 29

### DIFF
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -34,18 +34,20 @@ class SubscriptionsController < ApplicationController
   def verify
     return frequency_form_redirect unless valid_frequency
 
-    @back_url = url_for(
-      host: Plek.new.website_root,
-      action: :new,
-      topic_id: @topic_id,
-      frequency: @frequency,
-      address: @address,
-    )
-
     if @address.present?
+      @title = GdsApi.email_alert_api
+        .get_subscriber_list(slug: @topic_id)
+        .to_h.dig("subscriber_list", "title")
       VerifySubscriptionEmailService.call(@address, @frequency, @topic_id)
       render :check_email
     else
+      @back_url = url_for(
+        host: Plek.new.website_root,
+        action: :new,
+        topic_id: @topic_id,
+        frequency: @frequency,
+        address: @address,
+      )
       flash.now[:error] = t("subscriptions.new_address.missing_email")
       render :new_address
     end
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -3,9 +3,6 @@ class SubscriptionsController < ApplicationController
   before_action :assign_attributes
 
   def new
-    @title = GdsApi.email_alert_api
-      .get_subscriber_list(slug: @topic_id)
-      .to_h.dig("subscriber_list", "title")
     if @frequency.present?
       return frequency_form_redirect unless valid_frequency
 
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -3,15 +3,16 @@ class SubscriptionsController < ApplicationController
   before_action :assign_attributes
 
   def new
-    @back_url = @frequency.blank? ? govuk_url : url_for(action: :new, topic_id: @topic_id)
     @title = GdsApi.email_alert_api
       .get_subscriber_list(slug: @topic_id)
       .to_h.dig("subscriber_list", "title")
     if @frequency.present?
       return frequency_form_redirect unless valid_frequency
 
+      @back_url = new_subscription_path(topic_id: @topic_id)
       render :new_address
     else
+      @back_url = govuk_url
       render :new_frequency
     end
   end
@@ -26,7 +27,7 @@ class SubscriptionsController < ApplicationController
       )
     else
       flash.now[:error] = t("subscriptions.new_frequency.missing_frequency")
-      @back_url = url_for(action: :new, topic_id: @topic_id)
+      @back_url = new_subscription_path(topic_id: @topic_id)
       render :new_frequency
     end
   end
@@ -41,13 +42,7 @@ class SubscriptionsController < ApplicationController
       VerifySubscriptionEmailService.call(@address, @frequency, @topic_id)
       render :check_email
     else
-      @back_url = url_for(
-        host: Plek.new.website_root,
-        action: :new,
-        topic_id: @topic_id,
-        frequency: @frequency,
-        address: @address,
-      )
+      @back_url = new_subscription_path(topic_id: @topic_id, frequency: @frequency, address: @address)
       flash.now[:error] = t("subscriptions.new_address.missing_email")
       render :new_address
     end
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -33,9 +33,6 @@ class SubscriptionsController < ApplicationController
     return frequency_form_redirect unless valid_frequency
 
     if @address.present?
-      @title = GdsApi.email_alert_api
-        .get_subscriber_list(slug: @topic_id)
-        .to_h.dig("subscriber_list", "title")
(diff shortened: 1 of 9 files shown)


app/views/subscriptions/check_email.html.erb
```diff
@@ -1,20 +1,29 @@
...
+    <p class="govuk-body">
+      <%= t("subscriptions.check_email.description_html", address: @address) %>
+    </p>
+
+    <p class="govuk-body">
+      <%= t("subscriptions.check_email.instruction") %>
+    </p>
+
+    <p class="govuk-body">
+      <%= tag.strong @title %>
```
kevindew, line 16: Did you mean this to be `<strong><%= @title %></strong>` ?

config/locales/subscriptions.yml
```diff
@@ -21,7 +21,13 @@ en:
...
-      expiry: You have 7 days to confirm.
+      title: Check your email
+      description_html: "We’ve sent an email to <strong>‘%{address}’</strong>"
+      instruction: "Click the link in the email to confirm you want emails from GOV.UK about:"
+      expiry: The link will stop working after 7 days.
+      not_received_title: Not received an email?
+      not_received_detail_html: |
+        <p>Emails sometimes take a few minutes to arrive.</p>
+        <p>If you do not receive an email soon, check your spam or junk folder.</p>
+        <p>You may need to add gov.uk.email@notifications.service.gov.uk to your <a target="_blank" href="https://support.clickdimensions.com/hc/en-us/articles/115001160933-Add-a-Sender-to-the-Safe-Sender-List">safe sender list</a>.</p>
```
kevindew, line 24: Do we have precedent of using this link? It seems a pretty unusual destination to send people too, given that it's just some marketing company. I'd see if we could avoid the link and just have people Google it if we haven't got anything internal for this.

app/controllers/subscriptions_controller.rb
```diff
@@ -3,15 +3,16 @@ class SubscriptionsController < ApplicationController
...
 
   def new
-    @back_url = @frequency.blank? ? govuk_url : url_for(action: :new, topic_id: @topic_id)
     @title = GdsApi.email_alert_api
       .get_subscriber_list(slug: @topic_id)
       .to_h.dig("subscriber_list", "title")
     if @frequency.present?
       return frequency_form_redirect unless valid_frequency
 
+      @back_url = url_for(host: Plek.new.website_root, action: :new, topic_id: @topic_id)
```
kevindew, line 9: This shouldn't need to be an absolute url (and would likely cause problems in dev, flipping people onto the www hostname) - I think path_for would work? 

Was there a reason you added host? As I would assume it would work out the correct one based off the request.

app/controllers/subscriptions_controller.rb
```diff
@@ -3,15 +3,16 @@ class SubscriptionsController < ApplicationController
   before_action :assign_attributes
 
   def new
-    @back_url = @frequency.blank? ? govuk_url : url_for(action: :new, topic_id: @topic_id)
```
kevindew, line 6: Out of scope for today/this PR - but the whole `govuk_url` thing looks a bit fishy and ripe for a rethink.

app/controllers/subscriptions_controller.rb
```diff
@@ -3,9 +3,6 @@ class SubscriptionsController < ApplicationController
   before_action :assign_attributes
 
   def new
-    @title = GdsApi.email_alert_api
-      .get_subscriber_list(slug: @topic_id)
-      .to_h.dig("subscriber_list", "title")
```
kevindew, line 9: This also seems weirdly pointless due to: https://github.com/alphagov/email-alert-frontend/blob/1e866f29c1fd2aba8d278f38659cd4f2a24a0563/app/controllers/subscriptions_controller.rb#L61-L69

app/controllers/subscriptions_controller.rb
```diff
@@ -90,8 +90,6 @@ def govuk_url
       sanitised_referer_uri.query = referer_uri.query
       sanitised_referer_uri.fragment = referer_uri.fragment
       sanitised_referer_uri.to_s
-    elsif @subscriber_list["url"]&.match?(%r{^/get-ready-brexit-check/results})
-      (Plek.new.website_root + @subscriber_list["url"]).to_s
```
kevindew, line 94: Nice ⭐

app/controllers/subscriptions_controller.rb
```diff
@@ -33,9 +33,7 @@ def verify
     return frequency_form_redirect unless valid_frequency
 
     if @address.present?
-      @title = GdsApi.email_alert_api
-        .get_subscriber_list(slug: @topic_id)
-        .to_h.dig("subscriber_list", "title")
+      @title = @subscriber_list["title"]
```
kevindew, line 36: The commit doesn't seem to mention why this is moved outside of assign_attributes so I was wondering the reason for that?

Is this the only place we use `@title` - might it be simpler to reference `@subscriber_list["title"]` in the view?

kevindew (commented):
Nice job, I've got a few questions/comments.

config/locales/subscriptions.yml
benthorner: I did surprise me a bit. @gavinwye any thoughts?

app/controllers/subscriptions_controller.rb
benthorner: Agreed - I tidy this up in a later commit.

app/controllers/subscriptions_controller.rb
benthorner: Sure, fair enough. I've moved it to the view.

app/controllers/subscriptions_controller.rb
benthorner: I thought the same. I've got a follow-up item to think about it more.

app/controllers/subscriptions_controller.rb
benthorner: Oh yes, you're right! I was trying to be consistent with the other uses of "url_for" in the controller, on the assumption they do something special. I've rewritten the commit to remove it instead.

app/views/subscriptions/check_email.html.erb
benthorner: Yup - updated 👍

config/locales/subscriptions.yml
kevindew: Oh should the HTML here have govuk-frontend classes too: govuk-body and govuk-link ? or does the component have something to ensure these are styled correctly? It could be they look ok by default because of static but actually lack the styles they should have.

kevindew (commented):
Thanks - changes look good. Why we wait to hear about the link I perhaps caught another thing (I'm not certain it needs a change).

config/locales/subscriptions.yml
benthorner: Good point @kevindew 🕵️. It looks like it has [some of its own styles](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/details/_index.scss#L77). We also take [a similar approach in Content Publisher](https://github.com/alphagov/content-publisher/blob/master/app/views/components/_markdown_guidance.html.erb#L20).

benthorner: I've also removed the paragraph about safe senders, as discussed. @kevindew this should be ready for another review now.

kevindew: Nice digging thanks - not sure it's a great idea that we only style `<p>` elements. But alas, not todays problem.

kevindew (approved):

PR #918, opened 2020-11-24
Update start pages for new email signups

benthorner:

Please see the commits for more details. In order to keep the
design changes clear, I've split out refactorings into separate
commits e.g. for code that's been made redundant.

## Before

Screenshot 2020-11-24 at 14 09 55
Screenshot 2020-11-24 at 14 10 09
Screenshot 2020-11-24 at 14 10 25

## After

Screenshot 2020-11-24 at 14 05 08
Screenshot 2020-11-24 at 14 05 30
Screenshot 2020-11-24 at 14 07 48


app/views/content_item_signups/new.html.erb
```diff
@@ -22,32 +21,27 @@
                    track_category: "email_subscriptions",
                  }) do %>
 
+      <%
+         radio_items = @subscription.child_taxons.each.map do |taxon|
```
kevindew: Does this need `each.map` - I'm wondering if it's a mistake or a fruity data structure?

app/helpers/taxons_helper.rb
```diff
@@ -0,0 +1,11 @@
+module TaxonsHelper
```
kevindew, line 1: Might be that this is hinting towards content item being a model. That seems a later concern though and removing this presenter class looks a nice improvement.

kevindew (approved):
Looks good to me, potentially a minor thing to fix.

app/views/content_item_signups/new.html.erb
benthorner: Haha, no - not sure how that got in there. Removed 👍.

app/helpers/taxons_helper.rb
benthorner: Yeah, I did wonder about that. I think we need some more examples before modelling it, as the current usage is so specific to a particular document type.

app/views/content_item_signups/new.html.erb
kevindew: It looks like it was in the previous version, so more a c+p issue than you going mad 😁

kevindew (approved):

app/views/content_item_signups/confirm.html.erb
```diff
@@ -1,5 +1,4 @@
-<% page_heading = "Sign up to get emails" %>
-<% content_for :title, page_heading + ' - ' + @subscription.title %>
+<% content_for :title, t("content_item_signups.confirm.title") %>
 
 <% if live_content_item?(@content_item) %>
   <% content_for :back_link do %>
```
gavinwye, line 4: I thought we were removing the back link on all these pages?

app/views/content_item_signups/taxon.html.erb
```diff
@@ -0,0 +1,49 @@
+<% content_for :title, t("content_item_signups.taxon.title") %>
+
+<% if live_content_item?(@content_item) %>
+  <% content_for :back_link do %>
```
gavinwye, line 4: Remove?

app/views/content_item_signups/confirm.html.erb
benthorner: Indeed we are. That will come in a separate PR, as we made the decision half way through the work.

gavinwye:
✅ looks good to me 👍

PR #920, opened 2020-11-25
Remove back links for new subscription screens

benthorner:

This is a part of the new design for these screens. Note that the
functionality never fully worked:

- If the user advanced by one page, then we would lose any memory
of where they came from, and confusingly redirect them to "gov.uk".

- If the user came from an external site, we would confusingly
redirect them to "gov.uk".

While the new subscription flow is a multi-stage checkout that could
benefit from a "back" feature, this is still possible, and is more
reliable, via the native browser "back" functionality.


1pretz1 (approved):
I remember these backlinks didn't make much sense back 'int day so glad we're removing them! 👍

PR #921, opened 2020-11-25
Rewrite content for development homepage

benthorner:

Since a developer may not think to spin up the app in order to
learn about it, we should favour having documentation in the
README, which happens to already cover all of the content here.
On the other hand, keeping this page still provides a nicer
experience for someone who's new to the app. This replaces the
previous content with some simple links to further info.

## Before

Screenshot 2020-11-25 at 15 57 21

## After

Screenshot 2020-11-25 at 15 53 04

### DIFF
--- a/app/views/development/index.html.erb
+++ b/app/views/development/index.html.erb
@@ -3,52 +3,18 @@
 <%= render "govuk_publishing_components/components/title", title: "Email Alert Frontend" %>
 
 <p class="govuk-body">
-  This page is intended to only be shown in development and on Heroku review
-  apps. It won't be shown on GOV.UK as there are no routes to it.
+  This page is intended to only be shown in development.
 </p>
 
 <p class="govuk-body">
-  <%= link_to "Email Alert Frontend",
-              "https://github.com/alphagov/email-alert-frontend",
-              class: "govuk-link" %>
-  provides means to sign-up for GOV.UK mailing lists and manage
-  subscriptions.
+See the <%= link_to "README",
+            "https://github.com/alphagov/email-alert-frontend/blob/master/README.md",
+            class: "govuk-link" %> for usage.
 </p>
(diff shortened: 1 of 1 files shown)


1pretz1 (approved):
I think this was only added so govuk-docker wouldn't complain when starting the app up as it expected a root page to exist.

Your changes make sense regardless! 👍

benthorner:
Thanks for the review @1pretz1 👍.

> I think this was only added so govuk-docker wouldn't complain when starting the app up as it expected a root page to exist.

I don't think this is true, as GOV.UK Docker has no expectations of what an app does once it's started. Although you're right it's referenced in [the original commit](https://github.com/alphagov/email-alert-frontend/commit/1f3b1724abd42cdc8238f25cf0a2c82078dfc355#diff-c55cbc5bc88149a0b01407bd8c5cb31efe8ece59b583ff971aece1e70bff775a) - a @kevindew special 😄.

kevindew:
There used to be a check that queried [whether an app started up or not](https://github.com/alphagov/govuk-docker/blob/894457484abcb5abe01734ebeb4b7479bd876128/lib/govuk_docker/commands/startup.rb#L28) otherwise you got some red text: https://github.com/alphagov/govuk-docker/blob/894457484abcb5abe01734ebeb4b7479bd876128/lib/govuk_docker/commands/startup.rb#L36

Even though the GOV.UK Docker concern is gone, it's still nice to have this as it can be a disconcerting experience when you freshly start an app and get a 404 when you query it's root.

It sounds good to simplify the info.

benthorner:
> There used to be a check that queried whether an app started up or not otherwise you got some red text.

Oh yeah! Thanks for pointing that out @kevindew - I've tried to forget about the rather painful "Ruby CLI" era!

Issue #913, opened 2020-11-19
Issue moving through app


chris-gds:
There is an issue within this app as it's interacted with using the example page links

Starting 
`http://email-alert-frontend.dev.gov.uk/email-signup?link=%2Fgovernment%2Forganisations%2Fgovernment-digital-service`

As you move through the app the actual URL changes which generates an error page 
`http://www.dev.gov.uk/email/subscriptions/new?topic_id=government-digital-service`

![Screenshot 2020-11-19 at 13 51 45](https://user-images.githubusercontent.com/71266765/99674888-84571200-2a6e-11eb-87ab-1f6aadcc3d23.png)

![Screenshot 2020-11-19 at 13 51 53](https://user-images.githubusercontent.com/71266765/99674902-86b96c00-2a6e-11eb-80fc-c1b47b3b5304.png)

kevindew:
Thanks for raising @chris-gds 

I can see that is rather inconvenient in a dev environment. I had a look into it, hoping it would be an unnecessary absolute redirect that could be changed to relative, however it's a bit of a deeper pain than this. The URL comes from a different service, Email Alert API: https://github.com/alphagov/email-alert-api/blob/1711885b4df9c80aed5d8d2333e4e007c778badd/app/models/subscriber_list.rb#L49 - It'd probably be fine that this is a relative path, but not sure if that'd break things for any clients.

PR #922, opened 2020-11-26
Simplify request handling for content item signups

benthorner:

Fixes: https://github.com/alphagov/email-alert-frontend/issues/913

This PR makes a few changes to the the request handling:

- Use list "slug" instead of "subscription_url" for redirects
- Stop redirecting to "gov.uk" for blank redirect content items

I've also done a bit of refactoring, with the final part -
the controller tests - saved for a later PR, due to the
need to add some missing API test helpers first.

### DIFF
--- a/app/controllers/content_item_signups_controller.rb
+++ b/app/controllers/content_item_signups_controller.rb
@@ -6,7 +6,7 @@ class ContentItemSignupsController < ApplicationController
   include TaxonsHelper
 
   protect_from_forgery except: [:create]
-  before_action :validate_base_path
+  before_action :assign_content_item
   before_action :handle_redirects
   before_action :assign_list_params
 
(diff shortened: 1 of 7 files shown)


app/models/content_item_subscriber_list.rb
```diff
@@ -7,7 +9,8 @@ def initialize(content_item)
   end
 
   def subscription_management_url
```
kevindew: Should this be renamed subscription_management_path?

Although I actually wonder if this class is overstepping it's boundaries knowing the path, perhaps it could return the slug and the [code that calling it](https://github.com/alphagov/email-alert-frontend/blob/d00c3784fe5b1f7f937b593b8a84082849f879d8/app/controllers/content_item_signups_controller.rb#L26) should deal with the path. For example: `redirect_to new_subscription_path(topic_id: signup.topic_id)` ?

Also great stuff fixing this, if you weren't aware this should also resolve: https://github.com/alphagov/email-alert-frontend/issues/913

kevindew: Oh I see this change gets removed later in the PR making most of this comment moot.

app/controllers/content_item_signups_controller.rb
```diff
@@ -21,21 +21,15 @@ def new
...
 
-  PERMITTED_CONTENT_ITEMS = %w[taxon
-                               organisation
-                               ministerial_role
-                               person
-                               topic
-                               topical_event
-                               service_manual_topic
-                               service_manual_service_standard].freeze
-
```
kevindew, line 42: Nice job 👍

kevindew (approved):
Lots of nice improvements here 💯

app/models/content_item_subscriber_list.rb
benthorner: Oh nice, thanks for pointing out the issue - I've updated the PR description with it.

PR #928, opened 2020-12-04
Update links defined in yml

alex-ju:
Adds `govuk-link` classes to use the new styling and focus states and removes `target="blank"` as we don't seem to have a strong reason for using it.

Thanks @vanitabarrett for reporting this!


injms (approved):
🔗 👍

PR #933, opened 2020-12-08
Remove support for redirecting after sign in

benthorner:
https://trello.com/c/sgJIkObA/662-authenticate-before-unsubscribing-from-digest-emails

This isn't tested, and isn't used in practice: we always take the
user to their list of subscriptions.

### DIFF
--- a/app/controllers/subscriber_authentication_controller.rb
+++ b/app/controllers/subscriber_authentication_controller.rb
@@ -28,8 +28,7 @@ class SubscriberAuthenticationController < ApplicationController
     end
 
     authenticate_subscriber(token.data[:subscriber_id])
-    destination = safe_redirect_destination || list_subscriptions_path
-    redirect_to destination
+    redirect_to list_subscriptions_path
   end
 
 private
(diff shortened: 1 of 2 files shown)


kevindew (commented):
This seems good to me, nice removal, I don't have enough context to know whether it's not used at all though.

If I remember correctly the token is created in Email Alert API and is in this code: https://github.com/alphagov/email-alert-api/blob/528595124b00c199021e5693d81cb700d71a5a07/app/controllers/subscribers_auth_token_controller.rb#L20 so I'd have thought we'd have to remove the functionality there first.

benthorner:
> This seems good to me, nice removal, I don't have enough context to know whether it's not used at all though.

Sure, I almost forgot to do the other half of the removal, which should provide the rest of the context: https://github.com/alphagov/email-alert-api/pull/1502. In its commit, I've linked to evidence that it was only speculative.

In summary: it's never actually been used, so nothing depends on it being sent or received i.e. we can merge this PR and the one I linked to above in any order.

kevindew (approved):

PR #934, opened 2020-12-08
Require some kind of authentication to unsubscribe

benthorner:
https://trello.com/c/sgJIkObA/662-authenticate-before-unsubscribing-from-digest-emails

Please see the commits for more details. Note that the approach
changes in the final commit, as I realised the original way had
a flaw in it. Hopefully the story is easy to follow, but I can
try to rewrite the branch if not.

### DIFF
--- a/app/controllers/unsubscriptions_controller.rb
+++ b/app/controllers/unsubscriptions_controller.rb
@@ -49,11 +49,11 @@ private
   end
 
   def check_owns_subscription
-    return if authenticated? &&
-      @subscription.dig("subscriber", "id") == authenticated_subscriber_id
+    expected_id = @subscription.dig("subscriber", "id")
+    return if authenticated? && expected_id == authenticated_subscriber_id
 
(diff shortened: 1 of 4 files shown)


kevindew (approved):
This looks good to me. It doesn't seem too much of a pain if users caught in the transition have to sign in.

It did take me a little while to understand how check_owns_subscription worked by the way. I didn't really clock at first that the first couple of lines were users authorised via authentication and the second couple were authorised via token.

benthorner:
Thanks again @kevindew 🏅. I've added another commit with a couple of comments in the `check_owns_subscription` method, to try and clarify what it does. Do you have any alternative suggestions?

benthorner:
Tested and works on Integration ✅.

kevindew:
> Thanks again @kevindew 🏅. I've added another commit with a couple of comments in the `check_owns_subscription` method, to try and clarify what it does. Do you have any alternative suggestions?

Looks good, thanks 👍

PR #942, opened 2020-12-11
Record the origin of new content item-based lists

benthorner:
https://trello.com/c/3H67DU88/456-epic-record-the-origin-of-all-subscriber-lists-on-govuk

The URL property of subscriber lists was introduced to support the
Brexit Checker [1], but the functionality applies to all lists. We
think that recording these URLs will in future help us to determine
which lists are still "alive", by checking they refer to a live URL
on GOV.UK. This takes a step in that direction.

[1]: https://github.com/alphagov/finder-frontend/blob/2b565b6ccb6449d329dba00bb39d4d629a42c81a/app/controllers/brexit_checker_controller.rb#L189

### DIFF
--- a/app/services/generate_subscriber_list_params_service.rb
+++ b/app/services/generate_subscriber_list_params_service.rb
@@ -7,6 +7,7 @@ class GenerateSubscriberListParamsService < ApplicationService
     {
       "title" => content_item["title"],
       "links" => link_hash,
+      "url" => content_item["base_path"],
     }
   end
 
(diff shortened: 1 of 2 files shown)


kevindew (approved):
Looks good. 

Do I understand correct that this sign-up approach cannot be used for those that have a query string, for example: https://www.gov.uk/search/news-and-communications/email-signup?level_one_taxon=a544d48b-1e9e-47fb-b427-7a987c658c14&order=updated-newest and therefore we can be confident that for these the base_path truly represents the subscription and not just a parent that can be refined?

benthorner:
> Do I understand correct that this sign-up approach cannot be used for those that have a query string

Yes, that's [right](https://github.com/alphagov/email-alert-frontend/blob/34757972582f0c44a77386e87f9a0801d572c5dc/app/controllers/content_item_signups_controller.rb#L46). I might have a look at enabling this for new finder lists too.

benthorner:
Tested and works on Integration ✅.

PR #944, opened 2020-12-15
Replace list description with hard-coded text

benthorner:
https://trello.com/c/gbYVX8lK/690-send-users-a-link-to-their-results-as-part-of-subscribing-to-the-checker

Previously we relied on the list description being a markdown URL
for Brexit Checker results. However, we want to move away from a
freeform description in the API because:

- It's only used for Brexit Checker lists.
- It's freeform, so we can't track the links (in emails).
- The description is repetitive and inaccurate (really a URL).

This replaces the description with the same text, but hard-coded,
which makes it easier to replace, clearer in purpose, and (thus)
 easier to target for deletion in the future.

## Design (unchanged)

Screenshot 2020-12-15 at 14 41 26

### DIFF
--- a/app/views/subscriptions_management/index.html.erb
+++ b/app/views/subscriptions_management/index.html.erb
@@ -37,10 +37,12 @@
         <p class='govuk-body'>
           <%= subscription['created_at'].to_datetime.strftime("Created on %-d %B %Y at %-I:%M%P") %>
         </p>
-        <% if subscription['subscriber_list']['description'].present? %>
-          <%= render "govuk_publishing_components/components/govspeak", {} do %>
-            <%= raw(Kramdown::Document.new(subscription['subscriber_list']['description']).to_html) %>
-          <% end %>
+        <% if subscription['subscriber_list']['url'] =~ %r{transition-check/results} %>
+          <p class="govuk-body">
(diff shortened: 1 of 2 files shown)


app/views/subscriptions_management/index.html.erb
```diff
@@ -37,10 +37,12 @@
         <p class='govuk-body'>
           <%= subscription['created_at'].to_datetime.strftime("Created on %-d %B %Y at %-I:%M%P") %>
         </p>
-        <% if subscription['subscriber_list']['description'].present? %>
-          <%= render "govuk_publishing_components/components/govspeak", {} do %>
-            <%= raw(Kramdown::Document.new(subscription['subscriber_list']['description']).to_html) %>
-          <% end %>
+        <% if subscription['subscriber_list']['url'] =~ %r{transition-check/results} %>
```
hannako, line 40: Oh that's nice! TIL

hannako (approved):
👍

app/views/subscriptions_management/index.html.erb
benthorner: I'll assume you're talking about the groovy `%r` syntax, and not the overall hack 😄.

hannako: You assume correctly!

PR #954, opened 2020-12-29
Sign in after subscribing

benthorner:
https://trello.com/c/hdOrNhuC/669-sign-in-after-subscribing

This PR changes the sign up journey so that the user will see the
manage page, instead of a dead-end "you've subscribed" page. Please
see the commits for more details.

Note that the design of the flash message is a mixture of the old and
new content, in order to work with the existing "success" component -
we will update this uniformly in future work.

## Before

Screenshot 2020-12-29 at 13 15 39

## After

Screenshot 2021-01-07 at 17 43 09


app/controllers/subscriber_authentication_controller.rb
```diff
@@ -1,4 +1,6 @@
 class SubscriberAuthenticationController < ApplicationController
+  include SessionsHelper
```
kevindew, line 2: Minor aside comment - I'm never too sure whether helpers are meant to be used for sharing controller methods, particularly via the include method which has a different context to views. Your approach seems reasonable here as this seems to be a trodden path in this app, but I do wonder about what the idiomatic Rails way is.

I'm left wondering should helpers that have no view purpose be helpers or should they be concerns.

Rails 5 seems to have added new things to make helper usage in controllers cleaner - which seems to suggest they might be encouraged. There is now a [`helpers`](https://github.com/rails/rails/pull/24866) method which means no helpers need importing and all are available e.g `helpers.authenticate_subscriber`

Anyway aside over 🤔

config/locales/subscription_authentication.yml
```diff
@@ -4,8 +4,14 @@ en:
...
     authenticate:
-      message: "You’ve subscribed successfully"
+      message: "You’ve subscribed to GOV.UK emails"
       description:
-        immediately: "You’ll get an email from GOV.UK each time we add or update a page about: %{title}"
-        daily: "You’ll get one email a day from GOV.UK about: %{title}"
-        weekly: "You’ll get one email a week from GOV.UK about: %{title}"
+        immediately_html: |
+          <p>You’ll get an email from GOV.UK each time we add or update a page about:</p><br/>
+          <p>%{title}</p>
```
kevindew, line 11: By putting block level HTML into these I wonder if we're actually taking something that has a pattern and structure and smushing it into this one description field. 

Perhaps you've actually got a different type of flash message here that takes 3 parameters: the action title ("You've subscribed..."), the detail ("You'll get an email...") and the subject (The list title). It feels a bit like we've created a mini view in the locale file with this approach.

kevindew (commented):
Looks great - just a comment around the last commits approach to think about.

config/locales/subscription_authentication.yml
benthorner: I agree it has formatting, although structure might be going a bit far here, as I think the only reason for the newline is to call attention to the list title.

That said, and despite having seen this approach elsewhere, since the content needs to be tested, but is not really suitable for use in tests.

With your suggestion, we'd still need to manually inject a couple of `<br>` tags, which we could do in the controller and would also be a bit unusual, albeit perhaps a bit simpler overall. Is that what you were thinking of?

kevindew: > With your suggestion, we'd still need to manually inject a couple of <br> tags, which we could do in the controller and would also be a bit unusual, albeit perhaps a bit simpler overall. Is that what you were thinking of?

Nah, I was thinking more like the view being along the lines of:

```
      <% capture description do %>
         <p><%= flash[:success]["detail"] %></p><br><p><%= flash[:success]["subject"] %></p>
      <% end %>
      <%= render 'govuk_publishing_components/components/success_alert', {
        message: flash[:success]["message"],
        description: description
      } %>
```

Where the view understands the flash structure and knows how to convert it rather than this being a controller concern. Bit clunky I know just feels like view is the place to deal with the HTML structure.

benthorner: Ah, I wasn't so keen on that: conceptually I'm not sure "subject" means a whole lot. Looking at the other views, there's a precedent for having one flash per use case, so I've amended the (now) second-last commit to do that. It has the advantage that we're no longer passing around the list title, which avoids another instance of [this](https://trello.com/c/XoOHgGIh/473-fix-email-alert-frontend-raising-500-errors-when-users-unsubscribe-from-lists-with-long-titles).

What do you think?

(I also added another small commit, following a product review.)

app/views/subscriptions_management/index.html.erb
```diff
@@ -2,15 +2,30 @@
...
-
     <% if flash[:success] %>
       <%= render 'govuk_publishing_components/components/success_alert', {
         message: flash[:success]["message"],
         description: flash[:success]["description"],
       } %>
     <% end %>
 
+    <% if flash[:subscription] %>
+      <% subscription = @subscriptions[flash[:subscription]['id']] %>
```
kevindew, line 13: Is there any risk that this will result in a nil? I'm assuming flash data gets set on a previous request and there's always a remove chance that `@subscriptions` doesn't have the key - which, I guess, would then result in an exception?

You could resolve this with `if flash[:subscription] && (subscription = @subscriptions[flash[:subscription]['id']])`

kevindew (commented):
That's great - I think this looks much nicer handled in the view. Just a small concern to address.

app/views/subscriptions_management/index.html.erb
benthorner: Hmm...I don't think so, which makes it awkward to write a test for. This flash is only shown just after the subscription has been created. In theory, someone could unsubscribe from it in the small amount of time when the redirect is happening, but that's veering towards impossible. Can you think of another scenario?

kevindew: Hmm sure, I don't have a good example of a realistic scenario. It just feels like something where why risk a crash from my opinion - so an example of defensive coding thing rather than something to test. I don't mind too much if you want to leave the risk in but my suggestion would be to just implicitly check for it's existence before proceeding.

benthorner: OK. I've added another commit with a test for the subscription.

kevindew (approved):

PR #971, opened 2021-01-14
Require user to select taxon or a child taxon

benthorner:
https://trello.com/c/oYRFCtBm/668-iterate-error-content-and-behaviour-for-new-signup-workflow

This also starts to move towards using the "link" query param for email
signup. Please see the commits for more details.

## After

Screenshot 2021-01-14 at 10 52 38

### DIFF
--- a/app/controllers/content_item_signups_controller.rb
+++ b/app/controllers/content_item_signups_controller.rb
@@ -11,7 +11,7 @@ class ContentItemSignupsController < ApplicationController
   before_action :assign_list_params
 
   def new
-    if is_taxon?(@content_item) && taxon_children(@content_item).any?
+    if is_taxon_with_children?(@content_item)
       render "taxon"
     else
       render "confirm"
@@ -19,11 +19,9 @@ class ContentItemSignupsController < ApplicationController
   end
 
   def confirm
-    if is_taxon?(@content_item) && taxon_children(@content_item).any?
-      if params[:topic].nil?
-        flash[:error] = t("content_item_signups.taxon.no_selection")
-        render "taxon"
-      end
+    if is_taxon_with_children?(@content_item) && params[:topic].nil?
+      flash[:error] = t("content_item_signups.taxon.no_selection")
+      render "taxon"
     end
   end
 
--- a/app/controllers/content_item_signups_controller.rb
+++ b/app/controllers/content_item_signups_controller.rb
@@ -18,7 +18,14 @@ class ContentItemSignupsController < ApplicationController
     end
   end
 
-  def confirm; end
+  def confirm
+    if is_taxon?(@content_item) && taxon_children(@content_item).any?
+      if params[:topic].nil?
+        flash[:error] = t("content_item_signups.taxon.no_selection")
+        render "taxon"
+      end
+    end
+  end
 
   def create
     slug = GdsApi.email_alert_api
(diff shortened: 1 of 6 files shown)


app/controllers/content_item_signups_controller.rb
```diff
@@ -18,7 +18,14 @@ def new
     end
   end
 
-  def confirm; end
+  def confirm
+    if is_taxon?(@content_item) && taxon_children(@content_item).any?
```
sihugh, line 22: worth extracting this as `is_taxon_with_children?` ? I note it's used in a couple of places now, and it'll ensure that the logic stays in step

spec/features/content_item_signup_spec.rb
```diff
@@ -50,9 +50,7 @@ def when_i_visit_the_signup_page
   end
 
   def and_i_refine_my_selection
-    expect(page).to have_checked_field("topic-2")
```
sihugh, line 51: is it worth asserting that we have no checked fields?

app/controllers/content_item_signups_controller.rb
```diff
@@ -43,13 +43,11 @@ def handle_redirects
...
 
-    redirect_to(new_content_item_signup_path(topic: destination_path))
+    redirect_to(new_content_item_signup_path(link: destination_path))
   end
 
   def assign_content_item
-    # Topic param left in for backwards compatibility.
-    # Topic is the user-facing terminology for taxons. Expect the taxon base
-    # path to be provided in a param of this name.
+    # Note: the "topic" param has historically appeared in external links
```
sihugh, line 50: That's interesting. Do we get direct visits to signup pages from external sites?

spec/controllers/content_item_signups_controller_spec.rb
```diff
@@ -138,7 +138,7 @@ def make_request(params)
       stub_email_alert_api_does_not_have_subscriber_list("links" => { organisations: [content_id] })
       stub_email_alert_api_creates_subscriber_list("links" => { organisations: [content_id] }, "slug" => "my-list")
 
-      make_request(topic: "/my-organisation")
+      make_request(link: "/my-organisation")
```
sihugh, line 141: I note that organisation pages already use the `link=/organisation/path` form 👍

sihugh (commented):
A couple of small things, but looks good 👍

app/controllers/content_item_signups_controller.rb
benthorner: Yup. I don't actually know if such links make use of the "topic" query param, but I'm airing on the side of caution in case we change how taxon signup works in future, when it might be tempting to remove it here - whoever is making the change should go and check. I did also mean that our other apps use "topic". Once we've switched to "link", it should be much clearer in future analytics/logs whether this query param is actually used externally.

app/controllers/content_item_signups_controller.rb
benthorner: Sure, that makes sense. I've added another commit for that 👍.

spec/features/content_item_signup_spec.rb
benthorner: Oh yes, good point - I missed the controller spec doesn't actually cover the lack of pre-selection in the UI. I've amended the commit to negate this check.

benthorner:
Thanks for the review @sihugh! I think I've addressed all your comments, if you could take another look?

sihugh (approved):

PR #979, opened 2021-01-19
Fix component asset imports

andysellick:
- some assets were being included but weren't needed
- at least one (details) was needed (JS) but not included

(based this on the suggested Sass/JS functionality of the component guide, but also did some manual checking)


alex-ju (approved):
nice one! getting rid of `govspeak` is going to be great for performance and getting `details` in will improve browser coverage

PR #985, opened 2021-01-25
Update login screens to match new design

benthorner:
https://trello.com/c/l2cC0fY4/678-update-sign-in-journey-email-screens-to-match-design

This also refactors the related feature specs. Please see
the commits for more details.

## Before

Before | After
:---:|:---:
Screenshot 2021-01-25 at 15 41 00 | Screenshot 2021-01-25 at 15 42 13
Screenshot 2021-01-25 at 15 41 29  |  Screenshot 2021-01-25 at 15 42 38
Screenshot 2021-01-25 at 15 41 08 | Screenshot 2021-01-25 at 15 42 21
Screenshot 2021-01-25 at 15 41 18 | Screenshot 2021-01-25 at 15 42 29
Screenshot 2021-01-25 at 15 41 50 | Screenshot 2021-01-25 at 15 42 04


config/locales/subscriber_authentication.yml
```diff
@@ -18,6 +18,14 @@ en:
...
+        <p>We’ve sent an email to <strong>%{address}</strong></p>
+        <br>
+        <p>Click the link in the email to confirm your email address.</p>
+        <br>
+        <p>The link will stop working after 7 days.</p>
+      delay_title: Not received an email?
+      delay_html: |
+        <p>Emails sometimes take a few minutes to arrive.</p>
+        <p>If you do not receive an email soon, check your spam or junk folder.</p>
+        <p>If this does not work, <a href="https://www.gov.uk/contact/govuk">contact GOV.UK</a> for help.</p>
```
kevindew, line 31: I think you'll want class="govuk-link" on this link. Without it I recall links are almost the same as GOV.UK ones but with a subtle colour difference.

config/locales/subscriber_authentication.yml
```diff
@@ -18,6 +18,14 @@ en:
...
       heading: "Check your inbox"
-      confirmation: "We’ve sent an email to %{address}."
-      prompt: Click the link in the email to manage your subscriptions.
-      delay_warning: It can take up to 15 minutes for the email to arrive.
+      description_html: |
+        <p>We’ve sent an email to <strong>%{address}</strong></p>
+        <br>
+        <p>Click the link in the email to confirm your email address.</p>
+        <br>
+        <p>The link will stop working after 7 days.</p>
```
kevindew, line 26: This doesn't feel a semantic use of br - looks like it's being used to achieve styling. Is there a margin bottom modifier class that can be used instead?

kevindew (commented):
Nice one - looks a great improvement in the screenshots.

spec/features/manage_subscriptions_confirmation_email_spec.rb
```diff
@@ -22,7 +22,8 @@ def and_i_enter_my_email_address
       subscriber_id, @email_address
     )
 
-    fill_in I18n.t!("subscriber_authentication.sign_in.email_input.label"),
+    fill_in I18n.t!("subscriber_authentication.sign_in.email_input"),
+            name: :address,
```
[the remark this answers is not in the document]
benthorner: You're right the `name` option makes the text redundant. I left it in for consistency, but it's probably more confusing than helpful. I've removed it in the first commit.

config/locales/subscriber_authentication.yml
benthorner: Oh yes, I meant to fix that. Thanks for spotting. I've added the `no-visited-state` class, as the target is really an action page, rather than one for someone to read.

The `govuk-link` class appears to be unnecessary if we have the other one and are already in a body-like context, but I've added it in anyway as the two are always seen together.

config/locales/subscriber_authentication.yml
benthorner: Oh interesting. I'd always thought of `<br>` as a stylistic element, and the Internet jury is vague on this point. But I can see that semantically it makes more sense for it to only be used within a `<p>`.

I can see it's really GOV.UK confusing things here, by removing the blank lines that normal follow `<p>` elements. I've changed the 5th commit to apply the `govuk-body` class, which restores the spacing.

benthorner:
This should be good to go now - thanks for the insightful comments 💯. I've added a final commit to follow-up on https://github.com/alphagov/email-alert-frontend/pull/985#discussion_r564394788 in another place in the code - there's a small increase in vertical space, but essentially no change.

config/locales/subscriber_authentication.yml
kevindew: Super. Good shout on the no-visited-state. As I understand .govuk-link uses a different colour palette and is a slightly different blue. I presume the default one is the result of a legacy utility like govuk_frontend_toolkit.

kevindew (approved):

config/locales/subscriber_authentication.yml
benthorner: > I presume the default one is the result of a legacy utility like govuk_frontend_toolkit

Yeah, it does look like there's a colour state coming from our "core-layout", but I do also recall that we're trying to be less "global", or more specific, about where we apply styles these days.

Issue #914, opened 2020-11-19
Missing startup documentation


chris-gds:
## What?

Viewing this app for the first time and starting up in isolation (to resolve an a11y issue) it's not clear that the typical Frontend startup process [outlined in the docs](https://docs.publishing.service.gov.uk/) does not apply to this app. 

In order to **startup** this app and to be able to move through example pages and have them render, this app has to be started using Docker and the `--live` flag. 

`$ govuk-docker up email-alert-frontend-app-live`

## Why?
Without starting the up via this method, the app from a Frontend POV is unusable and you run into many errors.  
Eg you cannot use `/.startup.sh` or docker without live data.
 
[PR conversation with @kevindew around updating the README](https://github.com/alphagov/email-alert-frontend/pull/912) advised that it would be more desirable to get consensus with startup documentation across many apps and is most likely a wider topic. Updating docs as suggested above would be misleading and would age badly

kevindew:
Thanks for turning this into an issue @chris-gds 

> Viewing this app for the first time and starting up in isolation (to resolve an a11y issue) it's not clear that the typical Frontend startup process outlined in the docs does not apply to this app.

By this process do you mean: https://docs.publishing.service.gov.uk/manual/local-frontend-development.html ? and if so I'm not sure why it didn't apply as I think it should?

---

I think the thing that is a bit fishy is that very few GOV.UK app readme's have a "Running the application" the application section, and the lack of one suggests that someone forgot to add it, rather than it purposely isn't there. Indeed, it was [only recently removed](https://github.com/alphagov/email-alert-frontend/pull/766/commits/d6de99f6b1104f8e02a088f394ee41cf287c66f1) as the information was out of date.

If we're able to reach a resolution on this issue, it'd be good if it was something that could apply across various GOV.UK Docker apps so we'd want something simple that could be repeated.

Personally I'd be happy with something like:

```
### Running the application

This application is designed to be run within the [GOV.UK Docker](https://github.com/alphagov/govuk-docker) development environment.  You can also [run this application in isolation](https://docs.publishing.service.gov.uk/manual/local-frontend-development.html#using-startup-scripts) for frontend development.
```

I'd want to keep anything we put brief to reduce the risks of information falling out of date.

@benthorner, as someone involved in the previous removal, what are your thoughts?

benthorner:
Thanks for tagging me @kevindew. 

@chris-gds I'm nor sure I understand the problem. Could you help me understand what's wrong?

[...]

PR #990, opened 2021-02-05
Update README

chris-gds:
## What?

Update docs, providing more details about running the application using docker.

## Why? 

[Request to update docs](https://github.com/alphagov/govuk-ithc-documentation/issues/116) and to [close issue](https://github.com/alphagov/email-alert-frontend/issues/914).

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,10 @@ A frontend for creating and managing email subscriptions.
 
 ## Overview
 
+### Running the application
+
+This application is designed to be run within the [GOV.UK Docker](https://github.com/alphagov/govuk-docker) development environment it will be available on the host at http://email-alert-frontend.dev.gov.uk/.  You can run Email Alert Frontend against live, read-only data with `govuk-docker-up app-live`.
+
 ### Signup
 
 This app provides three routes for signing up to email:


README.md
```diff
@@ -4,6 +4,10 @@ A frontend for creating and managing email subscriptions.
 
 ## Overview
 
+### Running the application
+
+This application is designed to be run within the [GOV.UK Docker](https://github.com/alphagov/govuk-docker) development environment it will be available on the host at http://email-alert-frontend.dev.gov.uk/.  Using the `--live` flag can be used to test a read-only frontend app against live GOV.UK APIs.
```
1pretz1, line 9: I think some clarification around the `--live` flag is needed here as this flag is only available via the [startup script] rather than when using Docker. However, it does look like docker supports running the app in [live mode] too via `govuk-docker-up app-live`. Maybe we could update this last sentence so it just relates to Docker?

[startup script]: https://github.com/alphagov/email-alert-frontend/blob/master/startup.sh
[live mode]: https://github.com/alphagov/govuk-docker/blob/master/projects/email-alert-frontend/docker-compose.yml#L47-L59

chris-gds: Sure, what would you like the ideal copy to read? I can update.

1pretz1: Possibly something like:

```
You can run Email Alert Frontend against live, read-only data with `govuk-docker-up app-live`.
```

chris-gds: Ace, that's been updated. Hopefully that reads better?

1pretz1 (approved):

README.md
1pretz1: Yup, looks good to me! Thanks for updating.

chris-gds: Np, thanks for taking a gander and offering the ideal copy. 🙌🏼

chris-gds: Ahh I've realised I don't have auth on this one - would you be ok to the honours and click merge? @1pretz1

1pretz1: Ah, probably because Email Alert Frontend is continuously deployed and uses stricter merge permissions.

1pretz1: Sure thing, I like pressing big green buttons 😄

PR #1086, opened 2021-07-02
Various frontend changes for accounts

barrucadu:
Various frontend changes for account-based email management.  None of these words or screens have been reviewed by a designer, this is just so we can get something working up on staging (it's behind the feature flag) and then iterate.

---

[Trello card](https://trello.com/c/Vzd8so2T/874-implement-designs-for-managing-your-email-notifications-via-the-account)

### DIFF
--- a/app/controllers/subscriber_authentication_controller.rb
+++ b/app/controllers/subscriber_authentication_controller.rb
@@ -51,6 +51,8 @@ class SubscriberAuthenticationController < ApplicationController
     render :confirm_your_govuk_account
   end
 
+private
+
   helper_method def confirm_govuk_account_url
     ENV.fetch("GOVUK_ACCOUNT_CONFIRM_EMAIL_URL")
   end
@@ -58,8 +60,6 @@ class SubscriberAuthenticationController < ApplicationController
   helper_method def govuk_account_auth_enabled?
     ENV["FEATURE_FLAG_GOVUK_ACCOUNT"] == "enabled"
   end
-
-private
 
   def token
     @token ||= AuthToken.new(params.require(:token))
--- a/app/controllers/subscriber_authentication_controller.rb
+++ b/app/controllers/subscriber_authentication_controller.rb
@@ -39,7 +39,7 @@ class SubscriberAuthenticationController < ApplicationController
     head :not_found and return unless govuk_account_auth_enabled?
     reauthenticate_govuk_account and return if account_session_header.blank?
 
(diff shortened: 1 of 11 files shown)


config/locales/subscriber_authentication.yml
```diff
@@ -27,3 +27,9 @@ en:
         <p>Emails sometimes take a few minutes to arrive.</p>
         <p>If you do not receive an email soon, check your spam or junk folder.</p>
         <p>If this does not work, <a class="govuk-link govuk-link--no-visited-state" href="https://www.gov.uk/contact/govuk">contact GOV.UK</a> for help.</p>
+    use_your_govuk_account:
+      heading: "Sign in with your GOV.UK account"
+      description_html: |
+        <p class="govuk-body">You’ve set up a GOV.UK account for <strong>%{address}</strong></p>
```
huwd, line 37: I know we do it everywhere... but I stll feel like we need a better way of doing these paragraph chunks with embedded HTML.

It's particularly the wrapping in repeated `<p>` tags

But with that said I can't think up a better way of doing it 🤔 .
- Using individual locale strings into P tags in the HTML assumes that different languages would need three paragraphs to get across the same point (not impossible we might want to actually use these locales for languages at some point).
- We could make this a list, and have the HTML map over the top and inject it into HTML in the view or a helper. But then we've got nested <strong> and <a> tags here too, so we still end up with HTML in our strings.

I think it's especially the idea of having classes in our locale files that seem like it could be easy to miss if we want changes.

Anyway, just pondering out loud, no changes needed I think this is a fit with what we've done everywhere so far!

app/controllers/subscriber_authentication_controller.rb
```diff
@@ -51,6 +51,10 @@ def process_govuk_account
     render plain: "This GOV.UK account does not have a verified email address."
   end
 
+  helper_method def govuk_account_auth_enabled?
```
huwd, line 58: huh this is new syntax, I've not seen this before.

So is this equivalent to doing:
```
def govuk_account_auth_enabled?
...
end

helper_method: :govuk_account_auth_enabled?
```

In some ways I like the shortening, but it's a bit jarring first time I saw it...

Can this be a private method too?

barrucadu: Yeah, `def` returns the name of the method as a symbol, which is why this works.  I've seen both approaches in GOV.UK code.

I've made these private now.

huwd (approved):

PR #1106, opened 2021-08-05
Account layout for email management pages

danacotoran:
The accounts team are working towards the goal of enabling users who have linked their email notifications address and their GOVUK Account to manage their email subscriptions through their account.

Part of this work is making the email management pages (which live in `email-alert-frontend`) look the same as the account manager when the user is authenticated with a linked account.

This PR reorganises some of the views and requests the `gem_layout_account_manager` from static which will automatically apply account page furniture to the email management pages when the user is logged in with a linked account.

(if you're going to review this I would highly recommend the "Hide whitespace changes option)

### What it looks like now
Screenshot 2021-08-10 at 10 22 55

### What it will look like when someone's logged in with a linked account
Screenshot 2021-08-10 at 10 20 45

Dependent on: 
- https://github.com/alphagov/govuk_publishing_components/pull/2255
- https://github.com/alphagov/static/pull/2571

-------

https://trello.com/c/F0ZDY0RC

------

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -15,6 +15,8 @@ class ApplicationController < ActionController::Base
 
   helper_method :authenticated?
 
+  helper_method :use_govuk_account_layout?
+
   if ENV["BASIC_AUTH_USERNAME"]
     http_basic_authenticate_with(
       name: ENV.fetch("BASIC_AUTH_USERNAME"),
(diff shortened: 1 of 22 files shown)


app/controllers/application_controller.rb
```diff
@@ -26,6 +28,10 @@ def authenticated?
     session["authentication"].present?
   end
 
+  def govuk_account_layout?
+    defined?(use_govuk_account_layout?) && use_govuk_account_layout? == true
+  end
```
barrucadu, line 33: Is there an advantage of defining this `govuk_account_layout?` over having:

```ruby
def use_govuk_account_layout?
  false
end
```

here?

huwd (approved):
LGTM!

PR #1107, opened 2021-08-06
Things are changing

huwd:
[Trello](https://trello.com/c/WSFGT2rz/941-implement-basic-things-are-changing-screen-with-placeholder-content)

**Draft pending final content**

Add a new page, hidden behnd the account feature flag.
Creates a  page with:
- A header
- Some descriptive paragraphs
- A call to action button

The button varies it's content based on if the user is logged in or not.
In either case the user is directed to the same path, as it will start an authenticated request that will send the user down the sign in journey if they are not logged in.

### DIFF
--- a/app/helpers/account_helper.rb
+++ b/app/helpers/account_helper.rb
@@ -1,0 +1,5 @@
+module AccountHelper
+  def govuk_account_auth_enabled?
+    ENV["FEATURE_FLAG_GOVUK_ACCOUNT"] == "enabled"
+  end
+end
--- a/app/views/link_accounts/show.html.erb
+++ b/app/views/link_accounts/show.html.erb
@@ -1,0 +1,23 @@
+<% content_for :title, t("link_accounts.things_are_changing.heading") %>
+
+<div class="govuk-grid-row">
+    <%= render "govuk_publishing_components/components/heading", {
+      text: t("link_accounts.things_are_changing.heading"),
+      heading_level: 1,
+      font_size: "l",
+      margin_bottom: 6,
+    } %>
+  <%= sanitize(t("link_accounts.things_are_changing.description")) %>
+
+    <%= form_tag(process_govuk_account_path) do %>
+      <%= render "govuk_publishing_components/components/button", {
+        text: t("link_accounts.things_are_changing.action_button.#{@link_or_sign_in}"),
+        margin_bottom: true,
+        data_attributes: {
+          module: 'auto-track-event',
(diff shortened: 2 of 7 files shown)


spec/controllers/link_accounts_controller_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+        it "has the expected title" do
+          expect(html.title).to eq("#{I18n.t('link_accounts.things_are_changing.heading')} - GOV.UK")
+        end
+
+        it "has the expected heading" do
+          expect(html.css("h1").text).to eq(I18n.t("link_accounts.things_are_changing.heading"))
+        end
+
+        it "has a button that links to process_govuk_account_path" do
+          expect(html.at(".govuk-button")["href"]).to eq(process_govuk_account_path)
```
huwd, line 72: Wasn't actually sure where this link points to.
Having a snoop around the controller, assume `process_govuk_account_path` is the best one?

app/views/link_accounts/show.html.erb
```diff
@@ -0,0 +1,22 @@
...
+  <%= render 'govuk_publishing_components/components/button', {
+    text: t("link_accounts.things_are_changing.action_button.#{@link_or_sign_in}"),
+    href: process_govuk_account_path,
+    margin_bottom: true,
+    data_attributes: {
+      module: 'auto-track-event',
+      track_category: 'things_are_changing',
+      track_action: t("link_accounts.things_are_changing.tracking.action_button.#{@link_or_sign_in}"),
+    },
+  } %>
```
barrucadu, line 21: Let's change this to a POST form (and so add a `post` route for the controller method), so we get Rails' CSRF protection.

We still need to add some CSRF protection around authentication (eg, so you can't just send a user to gov.uk/email/manage/authenticate/account` and get their accounts merged) but that can be done in govuk_personalisation I think.

huwd: Ace, great change, thanks for suggesting.

That's now here in [this commit](https://github.com/alphagov/email-alert-frontend/pull/1107/commits/3af2e841f99a8afc66e1716bc53a622e9dbdcf62), let me know what you think?

app/helpers/account_helper.rb
```diff
@@ -0,0 +1,5 @@
+module AccountHelper
```
huwd, line 1: 🤔 wonder if some kind of shared feature flags should go in the personalisation Gem?

barrucadu (approved):

PR #1120, opened 2021-08-23
Clarify that you can only manage existing subscriptions here

barrucadu:
We frequently get support tickets along the lines of "I've entered my
email address on this page a dozen times and got not emails!  Your
system is preventing me from signing up to notifications on XYZ!"

But the problem is that you can't sign up to new notifications from
this page.  The text on this page doesn't say that, and the text on
the next page says to contact support.  So, users who find this page
first get frustrated.

So, really emphasise that this is for *existing* subscriptions, not
new ones.

---

Screenshot 2021-08-23 at 09 57 10

### DIFF
--- a/app/views/subscriber_authentication/sign_in.html.erb
+++ b/app/views/subscriber_authentication/sign_in.html.erb
@@ -28,7 +28,9 @@
   } %>
 <% end %>
 
-<p class="govuk-body"><%= t("subscriber_authentication.sign_in.description") %></p>
+<div class="govuk-body">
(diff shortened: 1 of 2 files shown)


bilbof (approved):
Fantastic change, thanks. It might be nice to link off to a topic page or a search page one can subscribe on, perhaps we also need a homepage for creating subscriptions.

benjamineskola (approved):

barrucadu:
I'm not sure what would be good pages to use as examples, so I'll leave that for now - hopefully this change just by itself will be enough to reduce the tickets we get.

PR #1123, opened 2021-08-25
Add Feature: Sign up with Accounts

huwd:
The provisional foundations for allowing sign up to a wider range of
notifications by content type, and enforcing the use of a GOV.UK Account
instead of creating a notifications specific account via email sign up.

The feature is currently gated behind a flag. We will need to add
finalised content before switching it on in production.

After this a follow up PR can remove code relating to the old sign up
routes.

---
[Trello](https://trello.com/c/HgavhWAU/940-implement-basic-account-y-email-subscription-sign-up-screen-with-placeholder-content)

### DIFF
--- a/spec/controllers/account_subscriptions_controller_spec.rb
+++ b/spec/controllers/account_subscriptions_controller_spec.rb
@@ -74,7 +74,6 @@ RSpec.describe AccountSubscriptionsController do
           stub_account_api_get_sign_in_url(
             redirect_path: new_account_subscription_path(base_path),
             level_of_authentication: "level0",
-            auth_uri: auth_uri,
           )
           get :new, params: { link: base_path }
         end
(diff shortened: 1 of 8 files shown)


spec/controllers/account_subscriptions_controller_spec.rb
```diff
@@ -0,0 +1,150 @@
...
+          response.should redirect_to(new_account_subscription_path({ link: base_path }))
+        end
+      end
+
+      context "when logged in" do
+        it "creates a subscriber list, links the subscription in accounts and redirects the user back to the content item" do
+          mock_logged_in_session("new-session-id")
+          stub_email_alert_api_creates_subscriber_list(
+            {
+              "title" => "Salary sacrifice",
```
huwd, line 137: I would have preferred to use a fixture declared in this file and then re-use it.
But other tests in this repo seemed to get content items from the [govuk-content-schemas](https://github.com/alphagov/govuk-content-schemas) repo, so I did the same.

I could load these from the [detailed guide](https://github.com/alphagov/govuk-content-schemas/blob/main/examples/detailed_guide/frontend/detailed_guide.json) example too, but seemed quick and easy to just type out the expectations here. The history on the file indicates they don't change often.

barrucadu:
Let's pause this until we have the final journeys

barrucadu:
We've built the feature outside of this PR.

PR #1131, opened 2021-09-17
Add bundle folder to .gitignore

alex9smith:
`git status` is unhappy about it and we already do this on [other repos](https://github.com/alphagov/account-api/blob/main/.gitignore#L8)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ log
 yarn-error.log
 public/assets
 coverage
+.bundle


.gitignore
```diff
@@ -4,3 +4,4 @@ log
 yarn-error.log
 public/assets
 coverage
+.bundle
```
barrucadu, line 7: Could you end the file with a newline?  I think most people's editors do that by default, so not doing so will cause some extra diff when someone next changes this.

alex9smith: Oops!

barrucadu (approved):

PR #1135, opened 2021-09-23
Use `govuk_personalisation` for change email link

alex9smith:
When a user is logged in to a GOV.UK account, we send them to the account
manager to manage their email address. Switch to getting this URL from
`govuk_personalisation` instead so it's consistent with other apps.

https://trello.com/c/u35RYZZW/1030-use-govukpersonalisation-for-change-email-link-in-email-alert-frontend

### DIFF
--- a/app/controllers/subscriptions_management_controller.rb
+++ b/app/controllers/subscriptions_management_controller.rb
@@ -99,7 +99,7 @@ private
 
   def set_account_change_email_url
     if session.dig("authentication", "linked_to_govuk_account")
-      @account_change_email_url = ENV.fetch("GOVUK_ACCOUNT_CHANGE_EMAIL_URL")
+      @account_change_email_url = GovukPersonalisation::Urls.manage
     end
   end
 
--- a/spec/controllers/subscriptions_management_controller_spec.rb
+++ b/spec/controllers/subscriptions_management_controller_spec.rb
@@ -105,7 +105,7 @@ RSpec.describe SubscriptionsManagementController do
 
     context "when the subscriber is logged in through a GOV.UK Account" do
       around do |example|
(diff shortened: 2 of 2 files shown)


app/controllers/subscriptions_management_controller.rb
```diff
@@ -1,6 +1,7 @@
 class SubscriptionsManagementController < ApplicationController
   include Slimmer::Headers
   include Slimmer::Template
+  include GovukPersonalisation::ControllerConcern
```
barrucadu, line 4: We don't need to pull in the personalisation controller concern here, as email-alert-frontend has its own session cookie - rather than replace that, we're just adding a flag to it when the user is logged in via an account.

alex9smith: Ah ok. I thought we needed to 'import' the module in some way for the controller to be able to see it, but it looks like Ruby is smarter than Python in that regard

huwd (approved):
LGTM

app/controllers/subscriptions_management_controller.rb
barrucadu: Rails is importing the gem automatically, and in the top-level module of the gem [we import all the modules](https://github.com/alphagov/govuk_personalisation/blob/main/lib/govuk_personalisation.rb).  An `import` in Ruby is less like a Python-style `import` and more like a C-style `#include` (kind of).

`include` merges a module into a different class / module, making all the things defined in the `include`d module available in the `include`ing one.

huwd: It's mostly "Rails magic" I think.
Rails monitors certain folders and filenames and autoloads them using "magic" (... which i think means metaprogramming)

PR #1143, opened 2021-09-30
Expose subscribers "not found" errors to the user

huwd:
[Trello](https://trello.com/c/P07i2NZV/1053-better-journey-for-users-who-dont-have-any-email-subscriptions)
[Figma](https://trello.com/c/P07i2NZV/1053-better-journey-for-users-who-dont-have-any-email-subscriptions)

## What is this?

Back in a previous commit [1] we introduced the idea that if a user
provides an email address that cannot be found in the subscribers
database we will carry on as usual and pretend to have sent a
verification email.

Whilst this is misleading to the user, the idea was that this protected
against using this form to repeatedly submit email addresses and find
out if a known email address had a GOV.UK account, or potentially to
trawl through iterations of email addresses to discover valid ones.

In this PR we revert that change and replace this with a message to
users explaining that the email address they provided cannot be found
and suggesting some actions.

The risk of information leakage remains, however we have decided that
the approach taken here is inconsistent with the wider approach we are
taking across GOV.UK accounts, which puts helping the user understand
ahead of the information leakage risk.

Our general rationale for this is that:
- This form is rate limited, protecting against the most crass brute
force attacks
- Information leaked in this way is likely to be very low value, only
showing that an email address uses a GOV.UK subscription but providing
no further detail about the nature or scope of the subscription
- The information that a user does not have an account is high value and
is likely to reduce helpdesk concerns and cases where uses accidentailly
misspell their address.

Under those circumstances this seems like an acceptable risk.

[1]: Commit ba1dd076fa9a3a2a3a4fab064dfa0a122aeb720b "Extract verifying
an email into services"

## What does it look like:
new_email_warning

## Where can i see it?

- Spin up this branch locally,
- visit `/email/manage/authenticate`
- enter an email address you know not to exist in the database
- see the screen


config/locales/subscriber_authentication.yml
```diff
@@ -40,3 +40,10 @@ en:
         <p class="govuk-body">You’ve set up a GOV.UK account for <strong>%{address}</strong></p>
         <p class="govuk-body">You now need to use your email address and password to manage your notifications.</p>
         <p class="govuk-body"><a class="govuk-link govuk-link--no-visited-state" href="%{process_govuk_account_path}">Sign in to your account</a>.</p>
+    no_subscriber_account:
```
barrucadu, line 43: Let's use `no_subscriber` instead of `no_subscriber_account` here (and elsewhere) - the term "subscriber account" isn't used anywhere else, we just talk about "subscribers".

barrucadu (approved):

PR #1156, opened 2021-10-19
Add a confirmation page for creating an account subscription

barrucadu:
This page handles signing up to notifications connected to the
account, which means it also handles the "linking" of an account-api
`OidcUser` to an email-alert-api `Subscriber`.

If a user has subscriptions which pre-date their account, we'll list
them so that the user knows what is changing.  The linking process
also triggers an email in this case (email-alert-api does this).

This page is intentionally very generic: it's not just for single-page
notifications, or content item subscriptions, it's for topics in
general.  This is so we can use this page across different sorts of
email journeys.

---

The frontend for this isn't quite right, the spacing around the button looks weird and the previous subscriptions don't look like a list:

Screenshot 2021-10-19 at 13 34 10
Screenshot 2021-10-19 at 13 37 51

---

[Trello card](https://trello.com/c/wUYJLuJ1/1064-implement-confirm-you-want-to-get-emails-page-account-linking-logic)

### DIFF
--- a/app/controllers/account_subscriptions_controller.rb
+++ b/app/controllers/account_subscriptions_controller.rb
@@ -1,0 +1,87 @@
+# frozen_string_literal: true
+
+class AccountSubscriptionsController < ApplicationController
+  include AccountHelper
+  include FrequenciesHelper
+  include GovukPersonalisation::ControllerConcern
+
+  DEFAULT_FREQUENCY = "daily"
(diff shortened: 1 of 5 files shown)


huwd (approved):

barrucadu:
Dropped the change the publishing task: it wasn't needed, as I put the pages under `/email/subscriptions`, which is a prefix route.

PR #1164, opened 2021-11-01
Fix "not subscribed" content

barrucadu:
This page doesn't match the designs: the email address is not
bold (and the link following it is), and the first sentence under the
"How to get emails about updates to GOV.UK" heading is missing.

I've also changed the `<div>`s to `<p>`s.

Screenshot 2021-11-01 at 11 05 49

---

[Trello card](https://trello.com/c/5ZoujfPL/1115-implement-email-management-account-sniffer-logic)


huwd (approved):
I'd be happy with that.
Should we ping @danacotoran for 👀 too?

PR #1165, opened 2021-11-01
Single page handler

huwd:
[Trello](https://trello.com/c/DsqzirDG/1065-implement-single-page-notifications-button-handling-page)
[Figma](https://www.figma.com/file/9RzsNFl3iYBXM5QfKO2HkE/Single-page-notifications-feature?node-id=1433%3A21852)

## What
Adds logic and tests for logic that signs users up to single page notifications.
Including a informational screen for logged out users.

## Why

We'll need to ensure users have a valid logged in session, with linked accounts before creating a subscription.
We can do those checks automatically for logged in users.

For logged out users we'll need to give them some information, and present a button that posts to DI's Single Sign on/up form, with the right params to return them and sign them up afterwards.

## What does it look like

image

### DIFF
--- a/app/controllers/single_page_subscriptions_controller.rb
+++ b/app/controllers/single_page_subscriptions_controller.rb
@@ -1,0 +1,52 @@
+class SinglePageSubscriptionsController < ApplicationController
+  include AccountHelper
+  include GovukPersonalisation::ControllerConcern
+
+  before_action do
+    head :not_found unless govuk_account_auth_enabled?
+  end
+
+  skip_before_action :verify_authenticity_token, only: [:show]
+
+  def show
+    content_item = GdsApi.content_store.content_item(topic).to_h
+    return unless logged_in?
+
+    subscriber_list = GdsApi.email_alert_api.find_or_create_subscriber_list(
+      {
+        url: topic,
+        name: content_item["title"],
+        content_id: content_item["content_id"],
+      },
+    ).to_h.fetch("subscriber_list")
+
+    result = CreateAccountSubscriptionService.call(subscriber_list, "daily", @account_session_header)
+    account_flash_add CreateAccountSubscriptionService::SUCCESS_FLASH
+    set_account_session_header(result[:govuk_account_session])
+
+    redirect_to topic
+  rescue GdsApi::ContentStore::ItemNotFound
+    head :not_found
+  rescue GdsApi::HTTPUnauthorized
+    logout!
+    redirect_with_analytics single_page_session_path
+  end
+
+  def edit
+    redirect_with_analytics single_page_session_path
+  end
+
+private
+
+  def topic
+    @topic ||= params.fetch(:topic)
+  end
+
+  def single_page_session_path
+    redirect_path = "#{confirm_account_subscription_path}?topic=#{topic}"
+
+    GdsApi.account_api.get_sign_in_url(
+      redirect_path: redirect_path,
+    )["auth_uri"]
+  end
+end
--- a/app/views/single_page_subscriptions/show.html.erb
+++ b/app/views/single_page_subscriptions/show.html.erb
@@ -1,0 +1,38 @@
+<% content_for :title, t("single_page_subscriptions.title") %>
+
+<%= render "govuk_publishing_components/components/heading", {
+  text: t("single_page_subscriptions.title"),
+  heading_level: 1,
+  font_size: "l",
+  margin_bottom: 6
+} %>
+
+<p class="govuk-body">
+  <%= t("single_page_subscriptions.account_required_description") %>
+</p>
+
+<%= render "govuk_publishing_components/components/heading", {
+  text: t("single_page_subscriptions.subheading"),
+  heading_level: 2,
+  font_size: "m",
+  margin_bottom: 6
+} %>
+
+<%= t("single_page_subscriptions.accounts_are_new_description_html") %>
+
+<%= render "govuk_publishing_components/components/inset_text", {
+} do
+  t("single_page_subscriptions.inset_description")
+end %>
+
+<%= t("single_page_subscriptions.usage_description_html") %>
+
+<%= form_tag single_page_new_session_path method: :post, :"data-module" => "explicit-cross-domain-links" do %>
+
+  <%= render "govuk_publishing_components/components/button", {
+    text: t("single_page_subscriptions.create_or_sign_in_description")
+  } %>
+
+  <%= hidden_field_tag(:topic, @topic) %>
+
+<% end %>
--- a/config/locales/single_page_subscriptions.yml
+++ b/config/locales/single_page_subscriptions.yml
@@ -1,0 +1,13 @@
+en:
+  single_page_subscriptions:
+    title: The way you subscribe to emails from GOV.UK is changing
+    account_required_description: You need a GOV.UK account to get these emails.
+    subheading: GOV.​UK accounts are new
+    accounts_are_new_description_html: |
+      <p class="govuk-body">They are a first step towards creating a single account where you can access all your government services.</p>
+      <p class="govuk-body">GOV.UK accounts will eventually replace all other government accounts.</p>
+    inset_description: At the moment, GOV.UK accounts are separate from other government accounts (for example Government Gateway or Universal Credit).
+    usage_description_html: |
+      <p class="govuk-body">Currently you can only use a GOV.UK account to manage your GOV.UK email subscriptions.</p>
+      <p class="govuk-body">We’ll be adding more features and services in the next few months.</p>
+    create_or_sign_in_description: Create a GOV.UK account or sign in
--- a/spec/controllers/single_page_subscriptions_controller_spec.rb
+++ b/spec/controllers/single_page_subscriptions_controller_spec.rb
@@ -1,0 +1,115 @@
+RSpec.describe SinglePageSubscriptionsController do
+  include GdsApi::TestHelpers::ContentStore
+  include GdsApi::TestHelpers::AccountApi
+  include GdsApi::TestHelpers::EmailAlertApi
+  include GovukPersonalisation::TestHelpers::Requests
+
+  render_views
+
+  let(:topic) { "/test" }
+  let(:topic_slug) { SecureRandom.uuid }
+  let(:topic_name) { "Test" }
+  let(:redirect_path) { "/email/subscriptions/account/confirm?topic=#{topic}" }
+  let(:auth_provider) { "http://auth/provider" }
+  let(:params) { { topic: topic } }
+
+  describe "when feature flag is not 'enabled'" do
+    it "POST /email/subscriptions/single-page/new-session returns 404" do
+      post :edit
+      expect(response).to have_http_status(:not_found)
+    end
+
+    it "POST /email/subscriptions/single-page/new" do
+      post :show
+      expect(response).to have_http_status(:not_found)
+    end
+  end
+
+  context "when the feature is on" do
+    around do |example|
+      ClimateControl.modify FEATURE_FLAG_GOVUK_ACCOUNT: "enabled" do
+        example.run
+      end
+    end
+
+    describe "POST /email/subscriptions/single-page/new-session" do
+      before { stub_account_api_get_sign_in_url(auth_uri: auth_provider, redirect_path: redirect_path) }
+
+      it "redirects to sign in with a redirect_path param" do
+        post :edit, params: params
+        expect(response).to redirect_to(auth_provider.to_s)
+      end
+
+      it "redirects with _ga param and cookie_consent if present in the request params" do
+        post :edit, params: params.merge({ _ga: "abc123", cookie_consent: "accept" })
+        expect(response).to redirect_to("#{auth_provider}?_ga=abc123&cookie_consent=accept")
+      end
+    end
+
+    describe "POST /email/subscriptions/single-page/new" do
+      before do
+        stub_content_store_has_item(topic)
+      end
+
+      it "404s when a content item can't be found" do
+        stub_content_store_does_not_have_item(topic)
+        get :show, params: params
+        expect(response).to have_http_status(:not_found)
+      end
+
+      context "when a user is not logged in" do
+        it "renders the view with a sign in link including the topic" do
+          get :show, params: params
+          expect(response.body).to include(single_page_new_session_path)
+        end
+      end
+
+      context "when a user is logged in" do
+        let(:session_id) { "session-id" }
+        let(:user_id) { "user-id" }
+        let(:subscriber_id) { "subscriber-id" }
+        let(:user_email) { "test@gov.uk" }
+        let(:subscription_list_id) { "subscription-list-id" }
+
+        before do
+          mock_logged_in_session(session_id)
+          stub_account_api_get_sign_in_url(auth_uri: auth_provider, redirect_path: redirect_path)
+
+          stub_email_alert_api_creates_subscriber_list({
+            url: topic,
+            title: topic_name,
+            slug: topic_slug,
+            id: subscription_list_id,
+          })
+
+          stub_email_alert_api_creates_a_subscription(
+            subscriber_list_id: subscription_list_id,
+            address: user_email,
+            frequency: "daily",
+            returned_subscription_id: "subscription-id",
+            skip_confirmation_email: false,
+            subscriber_id: subscriber_id,
+          )
+
+          stub_email_alert_api_link_subscriber_to_govuk_account(
+            session_id,
+            subscriber_id,
+            user_email,
+            govuk_account_id: user_id,
+          )
+        end
+
+        it "logs the user out if the session is invalid" do
+          stub_email_alert_api_link_subscriber_to_govuk_account_session_invalid(session_id)
+          post :show, params: params
+          expect(response).to redirect_to(auth_provider.to_s)
+        end
+
+        it "subscribes them and redirects back to the page" do
+          post :show, params: params
+          expect(response).to redirect_to("http://test.host#{topic}")
+        end
+      end
+    end
+  end
+end
--- a/app/views/development/index.html.erb
+++ b/app/views/development/index.html.erb
@@ -18,3 +18,24 @@ See the <%= link_to "README",
               govuk_publishing_components.component_guide_path,
               class: "govuk-link" %>.
 </p>
+
+<h3>Single page notifications</h3>
+
+  <p class="govuk-body">
+    The single page notification feature starts with a POST request to this app.
+    It then performs some conditional checks, and either subscribes the user or
+    shows them an inforamtional page and asks them to take further action.
+  </p>
+
+  <p class="govuk-body">
+    To test this page manually you can use the button below.
+  </p>
+
+  <%= form_tag "/email/subscriptions/single-page/new", method: :post do %>
+    <%= render "govuk_publishing_components/components/button", {
+      text: "Subscribe to single page notification"
+    } %>
+
+    <%= hidden_field_tag(:topic, "/guidance/keeping-a-pet-pig-or-micropig") %>
+
+  <% end %>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -36,7 +36,8 @@ Rails.application.routes.draw do
       post "/verify" => "subscriptions#verify", as: :verify_subscription
       post "/verify/account" => "subscriptions#verify_account", as: :verify_subscription_account
       get "/authenticate" => "subscription_authentication#authenticate", as: :confirm_subscription
-
+      post "/single-page/new" => "single_page_subscriptions#show"
+      post "/single-page/new-session" => "single_page_subscriptions#edit", as: :single_page_new_session
       scope "/account" do
         get "/confirm" => "account_subscriptions#confirm", as: :confirm_account_subscription
         post "/" => "account_subscriptions#create"


config/locales/single_page_subscriptions.yml
```diff
@@ -0,0 +1,14 @@
+en:
+  single_page_subscriptions:
+    title: The way you subscribe to emails from GOV.UK is changing
+    account_description_html: |
+      <p class="govuk-body">You need a new GOV.UK account to get these emails.</p>
+      <p class="govuk-body govuk-!-font-weight-bold">GOV.UK accounts are new</p>
+      <p class="govuk-body">They are a first step towards creating a single account where you can access all your government services.</p>
+      <p class="govuk-body">GOV.UK accounts will eventually replace all other governemnt accounts.</p>
```
barrucadu, line 8: In the Figma, it says "You need a GOV.UK account", not "You need a **new** GOV.UK account".

Should the "GOV.UK accounts are new" line be a heading?  If not we're kind of emulating a heading here with bold text, which doesn't feel great for accessibility.

config/locales/single_page_subscriptions.yml
```diff
@@ -0,0 +1,14 @@
...
+    inset_description_html: <p class="govuk-body">GOV.UK accounts are different from <a class="govuk-link" href="">other government accounts</a> you may have used before (for example Government Gateway or Universal Credit)</p>
```
barrucadu, line 9: Where should this link go?

Also, the Figma doesn't have "you may have used before"

app/helpers/account_helper.rb
```diff
@@ -2,4 +2,8 @@ module AccountHelper
   def govuk_account_auth_enabled?
     ENV["FEATURE_FLAG_GOVUK_ACCOUNT"] == "enabled"
   end
+
+  def single_page_notifications_feature_enabled?
+    ENV["SINGLE_PAGE_NOTIFICATIONS_FEATURE"] == "enabled"
+  end
```
barrucadu, line 8: I don't think we need a separate feature flag for this, as we're not going to release these things independently.

app/controllers/single_page_subscriptions_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+  before_action do
+    head :not_found unless single_page_notifications_feature_enabled?
+  end
+
+  def show
+    content_item = GdsApi.content_store.content_item(topic).to_h
+    if logged_in?
+      user_id = GdsApi.account_api.get_user(
+        govuk_account_session: @account_session_header,
+      ).to_h["id"]
```
barrucadu, line 14: `.to_h` isn't needed here

app/controllers/single_page_subscriptions_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+          govuk_account_id: user_id,
+        )
+      rescue GdsApi::HTTPNotFound
+        GdsApi.email_alert_api.link_subscriber_to_govuk_account(
+          govuk_account_session: @account_session_header,
+        )
+      ensure
+        subscriber_list_slug = GdsApi.email_alert_api.find_or_create_subscriber_list(
+          { url: topic, name: content_item["title"] },
+        ).to_h.dig("subscriber_list", "slug")
```
barrucadu, line 27: `to_h` isn't needed here.

Also, you need to pass the content ID.

app/controllers/single_page_subscriptions_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+    if logged_in?
+      user_id = GdsApi.account_api.get_user(
+        govuk_account_session: @account_session_header,
+      ).to_h["id"]
+
+      begin
+        GdsApi.email_alert_api.find_subscriber_by_govuk_account(
+          govuk_account_id: user_id,
+        )
+      rescue GdsApi::HTTPNotFound
```
barrucadu, line 20: You don't need this, you can just do

```ruby
GdsApi.email_alert_api.link_subscriber_to_govuk_account(
  govuk_account_session: @account_session_header,
)
```

app/controllers/single_page_subscriptions_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+      ensure
+        subscriber_list_slug = GdsApi.email_alert_api.find_or_create_subscriber_list(
+          { url: topic, name: content_item["title"] },
+        ).to_h.dig("subscriber_list", "slug")
+
+        GdsApi.account_api.put_email_subscription(
+          name: content_item["title"],
+          topic_slug: subscriber_list_slug,
+          govuk_account_session: @account_session_header,
+        )
```
barrucadu, line 33: Just call email-alert-api directly to create the subscription, don't go via the account-api here.

The email subscription functionality in account-api is so that we can have a subscription with a consistent name like "brexit checker" which may actually refer to different topics at different times.  But that's not the case here.

app/controllers/single_page_subscriptions_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+  def topic
+    @topic ||= params[:topic]
+  end
+
+  def single_page_session_path
+    redirect_path = "/email/subscriptions/account/confirm?topic=#{params[:topic]}"
+
+    GdsApi.account_api.get_sign_in_url(
+      redirect_path: redirect_path,
+    )["auth_uri"] + "?redirect_path=#{redirect_path}"
```
barrucadu, line 61: Why are you adding an extra `redirect_path` parameter here?

spec/controllers/single_page_subscriptions_controller_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+      it "redirects with _ga param and cookie_consent if present in the request params" do
+        post :edit, params: params.merge({ redirect_path: redirect_path, _ga: "abc123", cookie_consent: "accept" })
+        expect(response).to redirect_to("#{auth_provider}?redirect_path=#{redirect_path}&_ga=abc123&cookie_consent=accept")
+      end
+    end
+
+    describe "POST /email/subscriptions/single-page/new" do
+      before { stub_content_store_has_item(topic) }
+
+      it "404s when a content id cannot be found" do
```
barrucadu, line 52: Nitpick: it's when the page can't be found.

spec/controllers/single_page_subscriptions_controller_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+        end
+
+        it "logs the user out if the session is invalid" do
+          stub_account_api_unauthorized_user_info
+          post :show, params: params
+          expect(response).to redirect_to("#{auth_provider}?redirect_path=#{redirect_path}")
+        end
+
+        context "and they have no unlinked subscriptions" do
+          let(:subscriber_list_id) { SecureRandom.uuid }
```
barrucadu, line 94: This variable is unused

spec/controllers/single_page_subscriptions_controller_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+            stub_email_alert_api_find_subscriber_by_govuk_account(
+              user_id, subscriber_id, user_email
+            )
+            post :show, params: params
+            expect(response).to redirect_to("http://test.host#{topic}")
+          end
+        end
+
+        context "and they have unlinked subscriptions" do
+          let(:subscriber_id) { "subscriber-id" }
```
barrucadu, line 106: This is the same as the original value of the variable

spec/controllers/single_page_subscriptions_controller_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+              govuk_account_id: user_id,
+            )
+            post :show, params: params
+            expect(response).to redirect_to("http://test.host#{topic}")
+          end
+        end
+      end
+    end
+  end
+end
```
barrucadu, line 122: Nitpick: if you have a `context` it should have some `let`s or a `before` so it's clear what precondition it's describing.  Don't just put the precondition in the test body: it's easier to miss, and also won't work if you add a second test.

huwd: @alex9smith and I had a chat about this, what do you think Alex?
We talked a bit about doing it this way round vs in before blocks?

app/helpers/account_helper.rb
huwd: Do you mean just use: `FEATURE_FLAG_GOVUK_ACCOUNT` ?

app/controllers/single_page_subscriptions_controller.rb
huwd: We thought we'd need it so DI can return the user back here to try again after being logged in? Is that not the plan?

app/controllers/single_page_subscriptions_controller.rb
huwd: Would that be with [subscribe](https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/email_alert_api.rb#L76-L84) then?

So:
```
        GdsApi.account_api.subscribe(
          subscriber_list_id: subscriber_list_id,
          address:  "?" # guess we take this out of get_user call at the top?
        )
```

What values are we expecting for `frequency` and `skip_confirmation_email`?

app/controllers/single_page_subscriptions_controller.rb
huwd: Do you mean losing the entire `begin` block?

Is that because `link_subscriber_to_govuk_account` will return or link?

spec/controllers/single_page_subscriptions_controller_spec.rb
alex9smith: Yeah we went back and forth on this this for the last two tests. We could pull out the `stub_*` calls in each test to be a `before`:

```
context "and they have unlinked subscriptions" do
  before do
    stub_email_alert_api_find_subscriber_by_govuk_account_no_subscriber(user_id)
    stub_email_alert_api_link_subscriber_to_govuk_account(
      session_id,
      subscriber_id,
      user_email,
      govuk_account_id: user_id,
    )
  end

  it "merges their accounts, subscribes them to the topic and redirects back to the page" do
    post :show, params: params
    expect(response).to redirect_to("http://test.host#{topic}")
end
```

But then that leaves the contents of the `it` blocks the same for each test which we thought was a bit weird and confusing given they are testing different paths in the logic.

app/controllers/single_page_subscriptions_controller.rb
alex9smith: To clarify, we weren't sure what `get_sign_in_url` returned, and [the stub](https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/test_helpers/account_api.rb#L22-L29) implies it comes back with an `auth_uri` that doesn't have the redirect param, so we added it on here.

app/controllers/single_page_subscriptions_controller.rb
barrucadu: Yes, `find_subscriber_by_govuk_account` is only useful if you want to find the user *without* linking them.  If they're already linked, `link_subscriber_to_govuk_account` just returns them.

barrucadu: So, since you don't need to do `find_subscriber_by_govuk_account`, you also don't need to do `get_user`.

app/controllers/single_page_subscriptions_controller.rb
barrucadu: Use `GdsApi.email_alert_api`.  The call to `link_subscriber_to_govuk_account` above will return the subscriber, so you can get their address from that.

```ruby
subscriber_list = GdsApi.email_alert_api.find_or_create_subscriber_list(
  {
    url: topic,
    name: content_item["title"],
    content_id: content_item["content_id"],
  },
)

GdsApi.email_alert_api.subscribe(
      subscriber_list_id: subscriber_list.dig("subscriber_list", "id"),
      address: "take the address from link_subscriber_to_govuk_account",
      frequency: "daily",
      skip_confirmation_email: true,
    )
```

We've  been defaulting to daily subscriptions in other places, and the last email team wanted to phase out immediate subscriptions, so go with daily.

app/controllers/single_page_subscriptions_controller.rb
barrucadu: The stub comes back with whatever URI you want.  If you go to https://www.gov.uk you can see what URL you get redirected to to see what it's like:

https://oidc.account.gov.uk/authorize?client_id=LcueBVCnGZw-YFdTZ4S07XbQx7I&nonce=5cddc860bbe2d6ac6f753bfb8114de5d&redirect_uri=https%3A%2F%2Fwww.gov.uk%2Fsign-in%2Fcallback&response_type=code&scope=openid%20email%20govuk-account&state=1f2e57677c24b3458bc0dc1007dd50bf%3A312423&vtr=%5B%22Cl%22%2C%22Cl.Cm%22%5D

There's a `redirect_uri` parameter there which is what DI redirects us back to:

```
redirect_uri=https://www.gov.uk/sign-in/callback
```

Which is constructed from the `redirect_path` parameter in the `get_sign_in_url` call.

app/controllers/single_page_subscriptions_controller.rb
```diff
@@ -0,0 +1,58 @@
+class SinglePageSubscriptionsController < ApplicationController
+  include AccountHelper
+  include GovukPersonalisation::ControllerConcern
+
+  before_action do
+    head :not_found unless govuk_account_auth_enabled?
```
huwd, line 6: Is this right?
[Seemed that this was enabled in staging but not on integration or prod](https://github.com/alphagov/govuk-puppet/blob/501341433129483d2bbb7ed04cb3006332dcc6de/hieradata_aws/staging.yaml#L171)?

config/locales/single_page_subscriptions.yml
```diff
@@ -0,0 +1,13 @@
+en:
+  single_page_subscriptions:
+    title: The way you subscribe to emails from GOV.UK is changing
+    account_required_description: You need a GOV.UK account to get these emails.
+    subheading: GOV.​UK accounts are new
+    accounts_are_new_description_html: |
+      <p class="govuk-body">They are a first step towards creating a single account where you can access all your government services.</p>
+      <p class="govuk-body">GOV.UK accounts will eventually replace  all other government accounts.</p>
```
barrucadu, line 8: Nitpick: double space in `replace  all`

app/controllers/single_page_subscriptions_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def show
+    content_item = GdsApi.content_store.content_item(topic).to_h
+    if logged_in?
```
barrucadu, line 11: You could avoid the big indented block by changing this line to `return unless logged_in?`

app/controllers/single_page_subscriptions_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+private
+
+  def topic
+    @topic ||= params[:topic]
+  end
+
+  def single_page_session_path
+    redirect_path = "/email/subscriptions/account/confirm?topic=#{params[:topic]}"
```
barrucadu, line 52: Use a rails path helper here

app/controllers/single_page_subscriptions_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def edit
+    redirect_with_analytics single_page_session_path
+  end
+
+private
+
+  def topic
+    @topic ||= params[:topic]
```
barrucadu, line 48: If this parameter is required, should you use `fetch` here?

app/controllers/single_page_subscriptions_controller.rb
```diff
@@ -0,0 +1,58 @@
+class SinglePageSubscriptionsController < ApplicationController
```
barrucadu, line 1: I think you'll need to skip CSRF before `show`, as there won't be a CSRF token in the session.

spec/controllers/single_page_subscriptions_controller_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+        expect(response).to redirect_to("#{auth_provider}?_ga=abc123&cookie_consent=accept")
+      end
+    end
+
+    describe "POST /email/subscriptions/single-page/new" do
+      before do
+        stub_content_store_has_item(topic)
+      end
+
+      it "404s when a content id can't be found" do
```
barrucadu, line 54: Nitpick: it 404s when the content *item* can't be found.

app/controllers/single_page_subscriptions_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+        govuk_account_session: @account_session_header,
+      )
+
+      subscriber_list = GdsApi.email_alert_api.find_or_create_subscriber_list(
+        {
+          url: topic,
+          name: content_item["title"],
+          content_id: content_item["content_id"],
+        },
+      )
```
barrucadu, line 22: Does this need the parameters to be passed in a hash?  That's not the convention other things use.

app/controllers/single_page_subscriptions_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      GdsApi.email_alert_api.subscribe(
+        subscriber_list_id: subscriber_list.dig("subscriber_list", "id"),
+        address: subscriber.dig("subscriber", "address"),
+        frequency: "daily",
+        skip_confirmation_email: true,
+      )
+
+      redirect_to topic
```
barrucadu, line 31: You'll want to stick a success flash in the session here, so we can show it on the page.  There's a `AccountSubscriptionsController::SUCCESS_FLASH` constant you can use.

In #1170 I introduce a `CreateAccountSubscriptionService` to remove some repeated code around this sort of thing, but I can refactor this file to use it as part of that PR.

app/views/single_page_subscriptions/show.html.erb
```diff
@@ -0,0 +1,38 @@
...
+<%= t("single_page_subscriptions.accounts_are_new_description_html") %>
+
+<%= render "govuk_publishing_components/components/inset_text", {
+} do
+  t("single_page_subscriptions.inset_description")
+end %>
+
+<%= t("single_page_subscriptions.usage_description_html") %>
+
+<%= form_tag single_page_new_session_path method: :post do %>
```
barrucadu, line 30: You want the `explicit-cross-domain-links` on this form, to supply the parameters for the `redirect_with_analytics`

app/controllers/single_page_subscriptions_controller.rb
barrucadu: Ok, so now that #1170 is merged, you'll want to use `CreateAccountSubscriptionService` in this controller.

app/controllers/single_page_subscriptions_controller.rb
alex9smith: I think this is correct, unless I'm reading the [definition of `find_or_create_subscriber_list`](https://github.com/alphagov/gds-api-adapters/blob/91639289645173f92f4367f7d53f00e51957bc30/lib/gds_api/email_alert_api.rb#L9-L20) very wrong. 

It takes a single parameter `attributes` which is a hash that includes one of `content-id`, `links` or `tags`, but not more than one.

app/services/create_account_subscription_service.rb
```diff
@@ -26,6 +27,7 @@ def call
       subscriber_list_id: subscriber_list.fetch("id"),
       address: subscriber.fetch("address"),
       frequency: frequency,
+      skip_confirmation_email: @skip_confirmation_email,
```
barrucadu, line 30: Actually, I think this should have always had `skip_confirmation_email: true` - whoops.  So this doesn't need a separate `@skip_confirmation_email` parameter here.

alex9smith: Hmm yes, that makes sense. I'll drop the commit where I add it, sort this PR out then I'll open a new one that sets `skip_confirmation_email: true` in all cases and update the tests.

app/controllers/single_page_subscriptions_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+  def topic
+    @topic ||= params.fetch(:topic)
+  end
+
+  def single_page_session_path
+    redirect_path = "#{confirm_account_subscription_path}?topic=#{topic}"
+
+    GdsApi.account_api.get_sign_in_url(
+      redirect_path: redirect_path,
+    )["auth_uri"]
```
barrucadu, line 50: I think you should be able to construct this without the string interpolation, like so:

```ruby
    GdsApi.account_api.get_sign_in_url(
      redirect_path: confirm_account_subscription_path(topic: topic),
    )["auth_uri"]
```

spec/controllers/single_page_subscriptions_controller_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+      ClimateControl.modify FEATURE_FLAG_GOVUK_ACCOUNT: "enabled" do
+        example.run
+      end
+    end
+
+    describe "POST /email/subscriptions/single-page/new-session" do
+      before { stub_account_api_get_sign_in_url(auth_uri: auth_provider, redirect_path: redirect_path) }
+
+      it "redirects to sign in with a redirect_path param" do
+        post :edit, params: params.merge({ redirect_path: redirect_path })
```
barrucadu, line 39: Why are you passing the `redirect_path` as a parameter here?

spec/controllers/single_page_subscriptions_controller_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+        expect(response).to redirect_to(auth_provider.to_s)
+      end
+
+      it "redirects with _ga param and cookie_consent if present in the request params" do
+        post :edit, params: params.merge({ redirect_path: redirect_path, _ga: "abc123", cookie_consent: "accept" })
```
barrucadu, line 44: And here

spec/controllers/single_page_subscriptions_controller_spec.rb
alex9smith: Good spot, I think this is left over from some earlier testing

app/controllers/single_page_subscriptions_controller.rb
alex9smith: Doing it this way gives a webmock error where it's encoding the `topic` parameter. I get: 

```
Unregistered request: GET http://account-api.dev.gov.uk/api/oauth2/sign-in?mfa=false&redirect_path=/email/subscriptions/account/confirm?topic=%252Ftest

...


registered request stubs:

       stub_request(:get, "http://account-api.dev.gov.uk/api/oauth2/sign-in?mfa=false&redirect_path=/email/subscriptions/account/confirm?topic=/test")
```
I'm not sure how the route we're passing the user on to will cope with the encoded parameter, or how to set up the stub if it's ok.

barrucadu: Ok, this is probably fine then.  If something does need to be URL encoded to make it work, we'll discover that when we try things out in an actual web browser.

barrucadu (approved):
Looks good, go ahead and merge after squashing those fixup commits

PR #1168, opened 2021-11-02
Implement subscriber account authentication

barrucadu:
When a user comes to `/email/manage/authenticate`, they enter an email
address in a form.  We want to then:

- just log them in to email-alert-frontend, if they're logged into a
GOV.UK account and that GOV.UK account has the same email address; or

- tell them to sign in with their account, if there is a GOV.UK
account with the same email address but they're not logged into it; or

- fall back to the current magic link journey if the email address
does not belong to a GOV.UK account.

Screenshot 2021-11-02 at 12 42 47

---

[Trello card](https://trello.com/c/5ZoujfPL/1115-implement-email-management-account-sniffer-logic)

### DIFF
--- a/app/controllers/subscriber_authentication_controller.rb
+++ b/app/controllers/subscriber_authentication_controller.rb
@@ -14,10 +14,15 @@ class SubscriberAuthenticationController < ApplicationController
     end
 
     @address = params.require(:address)
-    VerifySubscriberEmailService.call(@address)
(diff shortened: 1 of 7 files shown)


huwd (commented):
Looks good, happy to approve when prerequisites are in and it's out of draft?

huwd (approved):

PR #1185, opened 2021-11-10
Reauthenticate user if they have no session

barrucadu:
`authenticate_subscriber_by_govuk_account` throws a 422 error if the
`govuk_account_session` parameter is not given, because it's
mandatory.

But the `rescue_from`s in this controller didn't handle that, and so
if an unauthenticated user got to this page somehow, they would be
served a 400 error.  This situation is unlikely (I only triggered it
by logging out and then manually visiting the URL) but still not
great.

So, gracefully handle this case, by checking if a session exists at
all before trying to use it.

---

[Trello card](https://trello.com/c/GKKcketU/1132-qa-snag-list)


alex9smith (approved):
Nice!

PR #1188, opened 2021-11-11
Do not create a session cookie for account users

barrucadu:
If a user is signing in to this app via their GOV.UK account, we don't
want to create a separate session cookie: if we do, we then have to
tie together deleting the email-alert-frontend session cookie when we
log a user out, and it's potentially confusing that the
email-alert-frontend session cookie will expire before the
`__Host-govuk_account_session` one

So, we change the authentication in this app to *preferentially* use
the GOV.UK account (see `authenticated_via_account?` ahead of
`session["authentication"}.present?` in the conditional).

When using account auth, no session cookie is created, we fall back to
linking with an email-alert-api subscriber on every request.  This
will be slower, but we can keep an eye on performance, as we expect a
gradual ramp-up of notifications users also using an account.

This does mean we have two totally separate authentication mechanisms
in the same app, until we migrate all notifications users over to the
GOV.UK account, which is a bit of a code smell.

---

[Trello card](https://trello.com/c/GKKcketU/1132-qa-snag-list)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,4 +1,7 @@
 class ApplicationController < ActionController::Base
+  include GovukPersonalisation::ControllerConcern
+  include AccountHelper
+
   include Slimmer::Template
 
   before_action :set_cache_control_header
@@ -14,7 +17,7 @@ class ApplicationController < ActionController::Base
   rescue_from GdsApi::HTTPGone, with: :gone
 
   helper_method :authenticated?
-
+  helper_method :authenticated_via_account?
   helper_method :use_govuk_account_layout?
 
   if ENV["BASIC_AUTH_USERNAME"]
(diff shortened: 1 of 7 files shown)


app/controllers/application_controller.rb
```diff
@@ -25,7 +28,22 @@ class ApplicationController < ActionController::Base
...
-    session["authentication"].present?
+    authenticated_via_account? || session["authentication"].present?
+  end
+
+  def authenticated_via_account?
+    return false unless govuk_account_auth_enabled?
+    return false if account_session_header.blank?
+    return @authenticated_via_account unless @authenticated_via_account.nil?
+
+    api_response = GdsApi.email_alert_api.link_subscriber_to_govuk_account(govuk_account_session: account_session_header)
```
alex9smith, line 39: I think it's a little confusing to see `link_subscriber_to_govuk_account` in a method that's called on every request. I understand why you've used it - anyone signed into the email alert frontend with their GOV.UK account should already have their accounts linked, so this will just return the subscriber object. 

I think `authenticate_subscriber_by_govuk_account` will do the same job here and be a lot clearer for future readers of this code. Plus we wouldn't run the risk of linking someone's accounts without informing them if there's an edge case we've not considered

barrucadu: > anyone signed into the email alert frontend with their GOV.UK account should already have their accounts linked, so this will just return the subscriber object.

Actually that's not the case, with this if you sign in to the email management frontend *without* having linked your account, it will be done in the background without you being asked if that's what you want.  Though you will get an email.

I raised this in our QA walkthrough, and this is desired behaviour: we think it will affect sufficiently few users that we don't need to make a separate screen for this.

alex9smith: Ah ok then 👍

alex9smith (approved):

PR #1189, opened 2021-11-12
Change the accounts feature flag to default to on

barrucadu:
**Do not merge until we are ready to go live!**

---

I've tried to keep the diff minimal: by leaving the feature flag in
place rather than just deleting it (to reduce the diff outside of the
tests), and by not collapsing now-unnecessary `context` blocks in the
tests.  But there's still quite a lot going on in this commit to make
the tests pass now.

A lot of the additions are new tests: I've added two new feature specs
for logging in and for subscribing with an account, and the
VerifySubscriptionEmailService didn't have any account-related tests
at all previously.

Later, when we delete the flag, we'll be able to simplify the tests a
fair bit.

---

[Trello card](https://trello.com/c/P2310odt/1134-make-single-page-notifications-live)


alex9smith (approved):
👍  Approving now so we can merge when we're ready to launch

barrucadu:
There will likely be a merge conflict with the tests in #1190 1190

PR #1190, opened 2021-11-12
Accounts frontend changes from QA session

barrucadu:
See commits for details.

But this isn't quite right currently:

1. The dividers between subscriptions aren't showing up.  There are `<hr>`s in the source (and were before I changed anything), so I guess some CSS is missing somewhere.
2. I don't know how to style the "topic" and "page" headers in the last commit to look like our designs, so I've just added them here to be styled.

Screenshot 2021-11-12 at 10 27 44

---

[Trello card](https://trello.com/c/4MKj7G4o/1135-govuk-account-email-subscriptions-page-design-amends)


alex9smith (commented):
The backend bits look good to me and good spot on the accessibility items @danacotoran! Have you got a screenshot of what it looks like now so we can share with the team and compare with Figma?

alex9smith (approved):

PR #1195, opened 2021-11-16
Update content on email management pages

alex9smith:
Make the heading a link for single page subscriptions and change 'created on' to 'you subscribed on' to match our designs.

[Trello](https://trello.com/c/Z9b4Ar5z/1156-update-email-subscriptions-management-page-content)
[Figma](https://www.figma.com/file/9RzsNFl3iYBXM5QfKO2HkE/Single-page-notifications-feature?node-id=4%3A2237)

There's a third item on the linked Trello card - the change for that will come in a separate PR.

The new page with a single page subscription:
image

### DIFF
--- a/app/views/subscriptions_management/index.html.erb
+++ b/app/views/subscriptions_management/index.html.erb
@@ -105,7 +105,7 @@
     <% end %>
 
     <p class='govuk-body'>
-      <%= subscription['created_at'].to_datetime.strftime("Created on %-d %B %Y at %-I:%M%P") %>
+      <%= subscription['created_at'].to_datetime.strftime(t("subscriptions_management.index.you_subscribed")) %>
     </p>
     <% if subscription['subscriber_list']['url'] =~ %r{transition-check/results} ||
       subscription['subscriber_list']['url'] =~ %r{get-ready-brexit-check/results} %>
--- a/app/views/subscriptions_management/index.html.erb
+++ b/app/views/subscriptions_management/index.html.erb
@@ -92,7 +92,7 @@
     </span>
 
     <%= render "govuk_publishing_components/components/heading", {
-      text: subscription['subscriber_list']['title'],
+      text: get_heading_content(subscription),
       heading_level: use_govuk_account_layout? ? 2 : 3,
       font_size: "m",
       margin_bottom: 4
--- a/app/helpers/subscriptions_helper.rb
+++ b/app/helpers/subscriptions_helper.rb
@@ -5,4 +5,12 @@ module SubscriptionsHelper
 
     content_item["public_updated_at"]
   end
+
+  def get_heading_content(subscription)
+    if subscription.dig("subscriber_list", "content_id").present?
+      link_to subscription["subscriber_list"]["title"], subscription["subscriber_list"]["url"], class: "govuk-link"
+    else
+      subscription["subscriber_list"]["title"]
+    end
+  end
 end
--- a/config/locales/subscriptions_management.yml
+++ b/config/locales/subscriptions_management.yml
@@ -16,6 +16,7 @@ en:
           daily: "You’ll get one email a day from GOV.UK about:"
           weekly: "You’ll get one email a week from GOV.UK about:"
       last_updated: "This page was last updated on %-d %B %Y"
+      you_subscribed: "You subscribed on %-d %B %Y at %-I:%M%P"
     no_subscriptions_warning_html: |
       <p class="govuk-body">You’re not currently getting emails about updates to GOV.UK.</p>
       <p class="govuk-body">Some GOV.UK pages have a link to ‘get emails’. You can use that link to subscribe to pages or topics you’re interested in.</p>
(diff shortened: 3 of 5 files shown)


app/views/subscriptions_management/index.html.erb
```diff
@@ -91,8 +91,16 @@
       <% end %>
     </span>
 
+    <% heading_text = capture do %>
```
huwd, line 94: humm variable assingment in a view always feels like a smell to me if we can avoid it...
but then one of the branches writes HTML...

@danacotoran, @alex9smith what do you think?

I'm wondering if we move this into the controller and set an instance variable like `@heading_text`, which either contains the `subscription['subscriber_list']['title']` or returns some HTML?

alex9smith: Yeah that's fair - I thought the same but then saw that there are loads of other variable assignments in this view and decided to be lazy 😅

alex9smith: I didn't want HTML in the controller or a variable in the view. Given the other variables in the view, this felt like the lesser of two evils to me

huwd: No I think it's fair, I had a similar thought at a glance.

However on closer inspection I think the other variable assignments here are actually doing something kind of cool.
They seem to be using capture blocks to create kind of [mini partials](https://github.com/alphagov/email-alert-frontend/blob/3849d857b1ee4a21908b400398401a2d69f6a90b/app/views/subscriptions_management/index.html.erb#L15-L18) that we can pass cleanly into our components.

I think what's different about what we're doing here is that we're doing an `if` statement within variable assignment

huwd: I think... then on the basis of complexity we might consider moving this to a helper?

Perhaps we could do something like:
```
def link_or_title(subscription)
  return { <a> ... etc } if is_single_page_subscription?(subscription)

  subscription['subscriber_list']['title']
end
```

Then call:

`link_or_title(subscription)` in the component?

huwd: though, now I think about it... it's a bit odd to be either returning HTML or a string...
Is that right? We've got a heading that can be a link here?

Let me go have an 👀 at the components

alex9smith: Yup, we want to have a heading that's a link if it's a single page subscription and text if not

huwd: I've had a look around, this seems cool from an accessibility standpoint, we seem to do it everywhere and it renders in the image you've provided, so ignore my momentary  🧠❓

alex9smith: I've refactored to use a helper in the controller for the logic. There will probably be a conflict with #1197 now, but it''l be easy enough to resolve.

app/controllers/subscriptions_management_controller.rb
```diff
@@ -118,4 +118,13 @@ def is_single_page_subscription?(subscription)
...
   helper_method :is_single_page_subscription?
+
+  def get_heading_content(subscription)
+    if is_single_page_subscription?(subscription)
+      "<a href='#{subscription['subscriber_list']['url']}' class='govuk-link'>#{subscription['subscriber_list']['title']}</a>"
+    else
+      subscription["subscriber_list"]["title"]
+    end
+  end
+  helper_method :get_heading_content
```
huwd, line 129: Yeah this looks good.
If we decide to go with moving this to a helper [over here](https://github.com/alphagov/email-alert-frontend/pull/1197#discussion_r751131532). Perhaps we could merge that one then rebase and let this one join it too?

In which case... I should test this, but I think because our method gets invoked from a view we should be able to make that line 124 use a `link_to` method (https://apidock.com/rails/ActionView/Helpers/UrlHelper/link_to).

The syntax on that can be a pain, but I think your class goes in the options hash.

If we did it that way... I have s suspicion we won't need to use santize... I think...

spec/views/subscriptions_management/index.html.erb_spec.rb
```diff
@@ -3,6 +3,7 @@
     assign(:subscriber, {})
     allow(view).to receive(:use_govuk_account_layout?).and_return(false)
     allow(view).to receive(:is_single_page_subscription?).and_return(false)
+    allow(view).to receive(:get_heading_content).and_return("A thing to subscribe to")
```
huwd, line 6: Another nice thing about helper filers is that they are super easy to unit test.
We can have a  `spec/helper/account_subscriptions_spec.rb` that  just pass a fake subscriber hash in with both bits we need to hit both code branches and cover that off as a simple unit test.

huwd (dismissed):
So dependent on the other one, happy to chat about these in slack again if you think I'm being over the top.

app/controllers/subscriptions_management_controller.rb
alex9smith: Agreed. I've moved the other one out into the helper so once that's in I'll rebase this branch and move this method out as well.

I'll have a look at `link_to` when I do that

alex9smith: Ok! I've done the refactor and `link_to` wasn't too bad either. 

I also took the opportunity to move the 'you subscribed at' into the locale file

barrucadu (approved):

PR #1197, opened 2021-11-16
Show last update date for single page subscriptions

alex9smith:
Since these subscriptions are for a single page, we can show the user
the last time they were updated so they can see if and when they would
have received an email about it.

This does call the content store once per single page subscription the
user has which could cause performance problems if a user has many
subscriptions. It should be fine for now but it's something we should
keep an eye on.

[Trello](https://trello.com/c/Z9b4Ar5z/1156-update-email-subscriptions-management-page-content)
[Figma](https://www.figma.com/file/9RzsNFl3iYBXM5QfKO2HkE/Single-page-notifications-feature?node-id=4%3A2237)

Screenshot with the new text: 
image

### DIFF
--- a/app/helpers/subscriptions_helper.rb
+++ b/app/helpers/subscriptions_helper.rb
@@ -1,0 +1,8 @@
+module SubscriptionsHelper
+  def get_updated_at_from_subscription(subscription)
+    base_path = subscription["subscriber_list"]["url"]
+    content_item = GdsApi.content_store.content_item(base_path)
+
+    content_item["public_updated_at"]
+  end
+end
--- a/app/views/subscriptions_management/index.html.erb
+++ b/app/views/subscriptions_management/index.html.erb
@@ -100,7 +100,7 @@
 
     <% if is_single_page_subscription?(subscription) %>
       <p class="govuk-body">
-        <%= get_updated_at_from_subscription(subscription).to_datetime.strftime("This page was last updated on %-d %B %Y") %>
+        <%= get_updated_at_from_subscription(subscription).to_datetime.strftime(t("subscriptions_management.index.last_updated")) %>
       </p>
     <% end %>
 
--- a/app/views/subscriptions_management/index.html.erb
+++ b/app/views/subscriptions_management/index.html.erb
@@ -98,6 +98,12 @@
       margin_bottom: 4
     } %>
 
+    <% if is_single_page_subscription?(subscription) %>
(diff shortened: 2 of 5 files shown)


app/controllers/subscriptions_management_controller.rb
```diff
@@ -118,4 +118,12 @@ def is_single_page_subscription?(subscription)
...
   end
   helper_method :is_single_page_subscription?
+
+  def get_updated_at_from_subscription(subscription)
+    base_path = subscription["subscriber_list"]["url"]
+    content_item = GdsApi.content_store.content_item(base_path)
+
+    content_item["public_updated_at"]
+  end
+  helper_method :get_updated_at_from_subscription
```
huwd, line 128: Interesting, so we're putting a lot of helper methods in the controller.
I think we can do that, it's useful when you want to use these methods as part of controller actions. So perhaps before you send a `@user` variable over to the view we use a method in the controller..

Then for whatever reason we might also want it to be separately available in the view. So we put it in here with that `helper_method` tag.

I _think_ if there's not a need for it to go in the controller, the correct place is in `app/helpers`.

These are cool, they're just really simple ruby modules ([our accounts one](https://github.com/alphagov/email-alert-frontend/blob/main/app/helpers/account_helper.rb) is a good example), that just define methods.

Rails magic then sweeps them up and automatically includes them into the views. It doesn't matter what we name these as [include_all_helpers defaults to true now](https://api.rubyonrails.org/classes/ActionController/Helpers.html)

So we could make something like:
```
# app/helpers/subscriptions_herlper.rb
module SubscriptionsHelper
  def get_updated_at_from_subscription(subscription)
    ...
  end
end
```

And it should automagically work in your view
And then yo

huwd (changes requested):
I'm tempted to say we should move those methods over to a helper?
Feel free to push back if you think that's OTT?

The locale bit is optional, if you have time / fancy it go for it, if not don't worry we can clean up views later?

app/controllers/subscriptions_management_controller.rb
alex9smith: I've moved this out into a helper. I'm not quite sure what the test should look like though

alex9smith: Ok, I've put in a simple helper test as well now. I've left the one in the controller as I think that's testing how we use the helper rather than the helper itself.

huwd (approved):

PR #1198, opened 2021-11-17
Fix single page unsubscriptions

alex9smith:
The key `subscriber_list_id` doesn't exist in the subscription hash, so
the unsubscribe path was never reached. Instead look for the ID in the
subscriber list object.

Also refactor the tests slightly so we assert the unsubscribe stub has
been called.

[Trello](https://trello.com/c/2NZWTfeo/1154-work-out-why-the-button-doesnt-unsubscribe-people-fix)


huwd (approved):
nice catch

PR #1199, opened 2021-11-17
RESTore sign up routes

huwd:
[Trello](https://trello.com/c/GptIhNbO/1160-refactor-single-page-notifications-controller-to-display-sign-in-page-using-two-requests)

We were fighting the RESTful framework when we wrote this.

In a perfect world the user is logged in and the initail POST request, goes and changes the state of their email subscriptions. However if they are not logged in we can't do that and we drop them out into an information screen with a request to log in or sign up.

Initially we had this screen being returned as a result of the POST request.
The more REST-y thing to do is to create a separate :show method on it's own GET route.

So that's what we're doing here.

Along the way I refactor out a bit of the topic_id stuff to safeguard against it not being there when we need it during :edit or :show journeys.

### DIFF
--- a/app/controllers/single_page_subscriptions_controller.rb
+++ b/app/controllers/single_page_subscriptions_controller.rb
@@ -10,6 +10,7 @@ class SinglePageSubscriptionsController < ApplicationController
 
   skip_before_action :verify_authenticity_token, only: [:create]
   before_action :fetch_subscriber_list, only: %i[create]
+  before_action :not_found_without_topic_id, only: %i[edit]
 
   def create
     return unless logged_in?
@@ -34,15 +35,22 @@ class SinglePageSubscriptionsController < ApplicationController
     redirect_to @content_item.fetch("base_path")
   rescue GdsApi::HTTPUnauthorized
     logout!
-    sign_in_and_confirm
+    sign_in_and_confirm(@topic_id)
   end
 
   def edit
-    @topic_id = params.fetch(:topic_id)
-    sign_in_and_confirm
+    sign_in_and_confirm(topic_id)
   end
 
 private
+
+  def not_found_without_topic_id
+    head :not_found and return unless topic_id
+  end
+
+  def topic_id
+    @topic_id = params[:topic_id]
+  end
 
   def fetch_subscriber_list
     @content_item = GdsApi.content_store.content_item(params.fetch(:base_path)).to_h
@@ -56,9 +64,9 @@ private
     @topic_id = @subscriber_list.fetch("slug")
   end
 
-  def sign_in_and_confirm
+  def sign_in_and_confirm(topic_id)
     redirect_with_analytics GdsApi.account_api.get_sign_in_url(
-      redirect_path: confirm_account_subscription_path(topic_id: @topic_id, return_to_url: true),
+      redirect_path: confirm_account_subscription_path(topic_id: topic_id, return_to_url: true),
     )["auth_uri"]
   end
 end
--- a/app/controllers/single_page_subscriptions_controller.rb
+++ b/app/controllers/single_page_subscriptions_controller.rb
@@ -8,11 +8,10 @@ class SinglePageSubscriptionsController < ApplicationController
     head :not_found unless govuk_account_auth_enabled?
   end
 
-  skip_before_action :verify_authenticity_token, only: [:show]
+  skip_before_action :verify_authenticity_token, only: [:create]
+  before_action :fetch_subscriber_list, only: %i[create]
 
-  before_action :fetch_subscriber_list, only: %i[show]
-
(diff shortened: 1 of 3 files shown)


spec/controllers/single_page_subscriptions_controller_spec.rb
```diff
@@ -151,11 +161,25 @@
...
+
+    describe "GET /email/subscriptions/single-page/new" do
+      let(:params) { { topic_id: topic_slug } }
+
+      it "returns 404 if no topic_id parameter is provided" do
+        get :show
+        expect(response).to have_http_status(:not_found)
+      end
+
+      it "returns 200 if a topic_id parameter is provier and renders an information page" do
```
alex9smith, line 179:
```suggestion
      it "returns 200 if a topic_id parameter is provided and renders an information page" do
```
nit pick - spelling

app/controllers/single_page_subscriptions_controller.rb
```diff
@@ -10,10 +10,13 @@ class SinglePageSubscriptionsController < ApplicationController
...
 
-  before_action :fetch_subscriber_list, only: %i[show]
+  before_action :fetch_subscriber_list, only: %i[create]
+  before_action :topic_id, only: %i[edit show]
 
-  def show
-    return unless logged_in?
+  def create
+    unless logged_in?
+      redirect_to new_single_page_subscription_path(topic_id: topic_id) and return
```
alex9smith, line 18: Could this be where the double render error is coming from? I'm not sure what Rails will try to do after it returns from here.

alex9smith (commented):
Looks good to me, this is a lot simpler than I was expecting so good work!

Not sure exactly where the double render error is coming from but happy to approve once you've sorted that.

huwd:
Oh just spotted my `main` is out of sync with `origin/main` will resolve that now... and work out what Jenkins is complaining about now.

huwd:
Oh i see I lost a line in the rebase, sorry - force push in 5

huwd:
Oh the double render was something really stupid, took me ages to work out what's going on but basically I got myslef into a trap.

I think it's an interesting one though, happy to take you through where I mucked up?
There's some interesting lessons for me next time on pre-mature memoisation, and trying to be too clever by half with instance variables and methods.

app/controllers/single_page_subscriptions_controller.rb
huwd: Yeah good spot, that wasn't it but this was a line that got lost in my rebase 🤦🏻

Check it out now, you'll see it goes properly to the redirect

huwd:
I've taken your spelling correction (thank you!!) and rebased it in.
I think the rest of the code survived my meddling this time round :)

app/controllers/single_page_subscriptions_controller.rb
```diff
@@ -35,16 +39,23 @@ def show
     redirect_to @content_item.fetch("base_path")
   rescue GdsApi::HTTPUnauthorized
     logout!
-    sign_in_and_confirm
+    sign_in_and_confirm(@topic_id)
```
alex9smith, line 42: Do you want `topic_id` instead of `@topic_id` here? Or is this one of those premature memoization things you mentioned

alex9smith (approved):
I don't think my question above is a blocking one, so I'll approve and we can talk about that separately if you like

app/controllers/single_page_subscriptions_controller.rb
huwd: Yeah so _this_ was the source of the bug.

Within the `create` method `@topic_id` is set by `fetch_subscriber_list`.
Because we are arriving with a base_path, then need to go get the topic_id out of the content store.

After this point we're sticking the topic_id into a URL param and so we can use `topic_id` that just reads things off the URL, and  because we're using `fetch` it lets us throw 404 if we try to visit `:edit` or `:show` without one.

This _is_ a bit confusing. Because now `@topic_id` and `topic_id` are getting set from two different sources... I am wondering if I should rename or separate these...

Anyway for these purposes, because we're in the `:create` method here, if we switch this to `topic_id` then it calls the private method, which tries to read it off the param, which isn't there... so it throws a param error... which rescues with a 404... which causes a double rending error 🤦🏻

huwd: I think you question is natural though, and may reveal that I've made a smell with overlapping names here...

Perahps I should change the name between topic_id we get out of content store and topic_id we start passing around as a param.

alex9smith: Right yes, I see. I think it does make sense when I look at the controller as a whole rather than just this diff. Renaming would help though, then I'm happy for you to merge after that

PR #1223, opened 2021-12-22
Handle signed out users who are already subscribed to a page

alex9smith:
Previously, if a user was signed out and clicked the 'get emails about
this page' button to subscribe to email alerts about a single page, we
didn't check if they already had a subscription to that page.

If they did we'd ask them to confirm again and send a new confirmation
email, but the manage emails page would show the date of when they
originally subscribed which was a bit confusing.

Now, do that check and if they're already subscribed redirect them back
to the content page after they sign in with a flash message to display
an info banner telling them they were already subscribed.

Note that this implementation does skip the confirmation page after signing 
in. We think this is ok - we're checking that the user is successfully logged 
in before we do anything (so we can't leak information) and we're not 
updating any information (because they already have the subscription) so 
it keeps the RESTful design of our app.

[Trello](https://trello.com/c/PkNG9Ecl/1190-journey-when-clicking-on-get-email-updates-when-i-have-a-subscription-already-but-am-not-signed-in)

### DIFF
--- a/app/controllers/account_subscriptions_controller.rb
+++ b/app/controllers/account_subscriptions_controller.rb
@@ -32,6 +32,8 @@ class AccountSubscriptionsController < ApplicationController
     set_account_session_header(response["govuk_account_session"])
     subscriber = response.to_h.fetch("subscriber")
 
+    redirect_with_already_subscribed if has_subscription?(subscriber)
+
     @address = subscriber.fetch("address")
 
     @unlinked_subscriptions =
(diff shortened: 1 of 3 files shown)


alex9smith:
Don't merge this before https://github.com/alphagov/government-frontend/pull/2322 otherwise we'll be redirecting users back to the content page but not telling them what happened.

huwd (approved):

app/controllers/account_subscriptions_controller.rb
```diff
@@ -74,4 +76,16 @@ def reauthenticate_user
       ),
     )["auth_uri"]
   end
+
+  def has_subscription?(subscriber)
+    subscriptions = GdsApi.email_alert_api.get_subscriptions(id: subscriber.fetch("id")).to_h.fetch("subscriptions", [])
```
barrucadu, line 81: We're now making two calls to this endpoint for users who aren't already subscribed

app/controllers/account_subscriptions_controller.rb
```diff
@@ -74,4 +76,16 @@ def reauthenticate_user
...
+
+    !subscriptions.find { |s| s["subscriber_list"]["id"] == @subscriber_list["id"] }.nil?
+  end
+
+  def redirect_with_already_subscribed
+    account_flash_add CreateAccountSubscriptionService::ALREADY_SUBSCRIBED_FLASH
+
+    redirect_to @subscriber_list["url"]
```
barrucadu, line 89: I think this should be `redirect_with_analytics`, to handle the case:

1. A user clicks the button on a page while logged out
2. They get redirected to DI to sign in
3. They get redirected to this endpoint from DI after signing in, with analytics parameters
4. We then redirect them back to the content page, and should preserve those analytics parameters because the javascript has not yet run

barrucadu:
I've opened #1226 to address my comments

PR #1226, opened 2021-12-29
Fix a couple of issues with the "already subscribed" feature

barrucadu:
See commit messages for more details.


alex9smith (approved):
Good spot

PR #1238, opened 2022-01-21
Show subscription name on the update frequency page

alex9smith:
Previously there was no mention of what subscription you're changing the
frequency of on this page which was confusing. Add the title from the
subscriber list so people know what they're changing.

Having `@subscriptions[@subscription_id]["subscriber_list"]["title"]` in
the view is a bit ugly, so add a new `SubscriptionsManagementHelper`
module to hold some new methods that are only used in the view.

[Trello](https://trello.com/c/ojh8IxlC/1223-changing-frequency-content-manage-page)

Screenshot of the new heading: 
image

### DIFF
--- a/app/helpers/subscriptions_management_helper.rb
+++ b/app/helpers/subscriptions_management_helper.rb
@@ -1,0 +1,5 @@
+module SubscriptionsManagementHelper
+  def get_subscription_title(subscription)
+    subscription["subscriber_list"]["title"]
+  end
+end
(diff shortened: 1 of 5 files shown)


alex9smith:
👍🏻  from the UCD team on the content

app/helpers/subscriptions_management_helper.rb
```diff
@@ -0,0 +1,9 @@
+module SubscriptionsManagementHelper
+  def get_subscription_title(subscription)
+    subscription["subscriber_list"]["title"]
+  end
+
+  def get_page_heading(subscription)
+    "How often do you want to get emails about ‘#{get_subscription_title(subscription)}’?"
```
barrucadu, line 7: I think this text should live in a locale file.  Locale entries can have variable interpolation, so the locale entry could be something like

```yaml
title: How often do you want to get emails about ‘%{subscription_title}’?
```

And then the view would use:

```ruby
page_heading = t("title", subscription_title: subscription["subscriber_list"]["title"])
```

alex9smith: Oh, that's nice. I'll update

barrucadu (approved):

PR #1256, opened 2022-02-16
Pass analytics params when signing in to manage page

alex9smith:
When an unauthenticated user tries to go to `/email/manage` they are
redirected through the sign in process. If they sign in with a GOV.UK
account this means they'll be taken off the `www.gov.uk` domain which
breaks the analytics tracking when they return as we weren't preserving
the linker parameters GOV.UK sign in are passing back to the app.

Now, pass through the parameters in the final redirect so they're
present on the page that Google Analytics loads on. Also add a test so
we don't accidentally break this in the future.

[Trello](https://trello.com/c/nNT2WjDH/1088-account-manager-tracking-and-reporting-update-after-di-migration)

### DIFF
--- a/app/controllers/subscriber_authentication_controller.rb
+++ b/app/controllers/subscriber_authentication_controller.rb
@@ -43,7 +43,7 @@ class SubscriberAuthenticationController < ApplicationController
 
   def process_govuk_account
     if authenticated_via_account?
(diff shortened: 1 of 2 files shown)


huwd (approved):
Nice and easy!

PR #1273, opened 2022-03-14
Add callout to interrupt page advising of mobile phone requirements

huwd:
## What
Add an advisory to the interrupt page - noting UK mobile phone requriement

## What it looks like
image


app/views/single_page_subscriptions/show.html.erb
```diff
@@ -45,6 +45,11 @@ end %>
 
 <%= form_tag single_page_new_session_path, method: :post, :"data-module" => "explicit-cross-domain-links" do %>
 
+  <%= render "govuk_publishing_components/components/inset_text", {
+  } do
```
danacotoran, line 49: Just curious are the empty `{ }` needed with this syntax?

huwd: lemme check, I'll admit this was a copy paste from futher up the doc

huwd: it still works! Have amended this commit and added another to remove from the other.

Still renders:
image

PR #1323, opened 2022-06-01
Allow AuthToken to accept SHA256 or SHA1 hashes

KludgeKML:

- Makes SHA256 tokens the default again (as per Rails 7 defaults)
- Adds a rotate to the decrypter that allows it to accept SHA1 tokens as well as SHA256 tokens.

This will allow it to accept the current SHA1 tokens from email-alert-api, and the SHA256 tokens from when email-alert-api is updated properly.

https://trello.com/c/Iz4TyAzx/1321-fixes-for-email-alert-api-email-alert-frontend-update-incident

Note - due to the way email-alert-api is set up, not currently possible to test this end to end. Has been tested by generating a token in the email-alert-api integration console:

`token = AuthTokenGeneratorService.call(hello: 'world')`

...then checking it decrypts on the email-alert-frontend integration console:

```
irb(main):007:0> token = AuthToken.new(data)
=> #<AuthToken:0x00007f23de69fb78 @token="1xZjrFbz4korhQywWP9H2nycQBNdySnzq...
irb(main):008:0> token.valid?
=> true
irb(main):009:0> token.data
=> {:hello=>"world"}
```

### DIFF
--- a/spec/models/auth_token_spec.rb
+++ b/spec/models/auth_token_spec.rb
@@ -2,14 +2,14 @@ describe AuthToken do
   include TokenHelper
 
   describe "#data" do
-    it "returns the data hash when valid" do
-      token = AuthToken.new(encrypt_and_sign_token(data: { a: "b" }))
+    it "returns the data hash when given a valid SHA1 token" do
+      token = AuthToken.new(encrypt_and_sign_token(data: { a: "b" }, hash_digest_class: OpenSSL::Digest::SHA1))
       expect(token.data).to eq(a: "b")
     end
 
-    it "returns nil after the expiry time" do
-      token = AuthToken.new(encrypt_and_sign_token(expiry: 0))
-      expect(token.data).to be_nil
+    it "returns the data hash when given a valid SHA256 token" do
+      token = AuthToken.new(encrypt_and_sign_token(data: { a: "b" }, hash_digest_class: OpenSSL::Digest::SHA256))
+      expect(token.data).to eq(a: "b")
     end
 
     it "returns nil if the token is malformed" do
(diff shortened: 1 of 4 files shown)


spec/models/auth_token_spec.rb
```diff
@@ -2,14 +2,28 @@
...
-      token = AuthToken.new(encrypt_and_sign_token(data: { a: "b" }))
-      expect(token.data).to eq(a: "b")
+    context "with a SHA1 token" do
+      it "returns the data hash when valid" do
+        token = AuthToken.new(encrypt_and_sign_token(data: { a: "b" }, hash_digest_class: OpenSSL::Digest::SHA1))
+        expect(token.data).to eq(a: "b")
+      end
+
+      it "returns nil after the expiry time" do
+        token = AuthToken.new(encrypt_and_sign_token(expiry: 0, hash_digest_class: OpenSSL::Digest::SHA1))
```
ChrisBAshton, line 12: These two `returns nil` tests don't give me much confidence as they don't pass the `data: { a: "b" }` part of the hash, so even without an `expiry: 0`, I'd expect `token.data` to be `nil` 🤔 
Could you add the `data` property to explicitly test this?

ChrisBAshton (commented):
Well done for coming up with a solution to this thorny problem!
Just so I'm clear, we'd expect to be able to revert this PR within say an hour of updating email-alert-api, right?

Nitpick: could you copy all the useful PR description into the commit message itself? Saves a future developer having to navigate to the PR in Git if their local IDE shows git commit information.

KludgeKML:
Reverting - the links are advertised as being valid for 7 days, so I think we'd want to wait a week between updating email-alert-api and updating this (it wouldn't strictly be a revert because currently email-alert-frontend is only accepting SHA1 tokens, and we'd want it to move to only accepting SHA256 once we were sure all the SHA1 tokens from email-alert-api were expired).

Details in commit - yeah, I can update the commit with the notes.

spec/models/auth_token_spec.rb
KludgeKML: The default data is actually an empty hash, so if you set the expiry to (say) 1000, or remove it entirely, the test will fail - it's only nil if the token is invalid for some reason. BUT! there's actually a nicer way of doing this - I can expand the test suite and move these tests into a test for the #valid? method, which would be a bit more explicit.

KludgeKML: Have updated the tests, see if the current change seems less worrying.

spec/models/auth_token_spec.rb
```diff
@@ -2,14 +2,18 @@
   include TokenHelper
 
   describe "#data" do
-    it "returns the data hash when valid" do
-      token = AuthToken.new(encrypt_and_sign_token(data: { a: "b" }))
-      expect(token.data).to eq(a: "b")
+    context "with a SHA1 token" do
+      it "returns the data hash when valid" do
```
ChrisBAshton, line 6: Nitpick: we tend to avoid using `context` blocks unless there's more than one `it` block inside it. In this case, we'd usually just do:

```suggestion
    it "returns the data hash when given a valid SHA1 token" do
```

ChrisBAshton (approved):
LGTM 👍 probably worth building the branch to the relevant env to check it works, before merging.

One nitpick comment but doesn't block the PR.

ChrisBAshton (approved):

PR #1423, opened 2022-11-03
Prevent Bad URI errors from getting into Sentry

KludgeKML:
Since we're not doing anything about these (they're just background noise of hacking attempts), we catch these errors and return 422s instead of crashing. To make sure we can still monitor if we want to, add the error details as logged warnings.

This also pulls the URL parsing that's done in the content store API call forwards into email-alert-frontend so we protect content store from bogus calls.

https://trello.com/c/Wd2RY7tJ/1632-silence-bad-uri-errors-on-email-alert-frontend


kevindew (approved):
Looks good to me, thanks for this ⭐

PR #1470, opened 2023-01-12
Dockerfile cleanup + enable bootsnap.

sengi:
- Use bootsnap to reduce startup resource usage spike.
- Parameterise Ruby version.
- Make better use of WORKDIR.
- Use env vars from the base image where appropriate, instead of hardcoding paths.
- Remove unnecessary use of `bundle exec`.
- Add .dockerignore.

Tested: app boots successfully with `docker run`.


Dockerfile
```diff
@@ -1,32 +1,26 @@
...
+WORKDIR $APP_HOME
+COPY Gemfile* .ruby-version ./
 RUN bundle install
-
-COPY . /app
-# TODO: We probably don't want assets in the image; remove this once we have a proper deployment process which uploads to (e.g.) S3.
-RUN bundle exec rails assets:precompile && rm -fr /app/log
+COPY . .
+RUN bootsnap precompile --gemfile .
+RUN rails assets:precompile && rm -fr log
```
theseanything, line 13: is the TODO comment above still valid?

sengi: Good question :) My take: the question in the TODO is still valid, but I decided not to keep it in here because 1) not sure it's all that helpful having a TODO repeated across 50+ repos and 2) I'm leaning more towards keeping the assets in cos it makes the images quite a bit more flexible (e.g. local e2e testing with minikube/kind/k3s/whatever, makes it simply if we want the option to serve everything from origin as a CDN-is-down [black swan](https://en.wikipedia.org/wiki/The_Black_Swan:_The_Impact_of_the_Highly_Improbable) contingency plan, etc.) and it doesn't add all that much to image size or build time and keeps release and rollout simple and decoupled. Could still be persuaded either way though.

PR #1570, opened 2023-07-13
Adds handler for from=your-services param hint in /email/manage

KludgeKML:
In order to improve the logged-in experience, in https://github.com/alphagov/frontend/pull/3670 we are redirecting /account directly to home.account.gov.uk. This means that we are skipping session creation. It improves the general logged in behaviour, but means that if someone goes to /account and logs in, then follows the link to /email/manage, email-alert-frontend will not know that the user is logged in (because no session will exist), and they'll be prompted for their email address. To get around this, we add support for a hint parameter (from=your-services) which will be added to the link in the home.account.gov.uk/your-services page. When we go to /email/manage?from=your-services, the app knows that we came from One Login and are therefore probably logged in, so attempts a silent login.

https://trello.com/c/SiAQ3dZU/2086-redirect-govuk-account-to-homeaccountgovuk

### DIFF
--- a/app/controllers/subscriptions_management_controller.rb
+++ b/app/controllers/subscriptions_management_controller.rb
@@ -1,6 +1,7 @@
 class SubscriptionsManagementController < ApplicationController
   include Slimmer::Headers
   include Slimmer::Template
+  before_action :handle_one_login_hint, only: [:index]
   before_action :require_authentication
   before_action :get_subscription_details
   before_action :set_back_url
@@ -100,6 +101,12 @@ class SubscriptionsManagementController < ApplicationController
   end
 
 private
+
+  def handle_one_login_hint
+    return unless params[:from] == "your-services" && !authenticated?
+
+    redirect_with_analytics GdsApi.account_api.get_sign_in_url(redirect_path: list_subscriptions_path)["auth_uri"]
+  end
 
   def get_subscription_details
     subscription_details = GdsApi.email_alert_api.get_subscriptions(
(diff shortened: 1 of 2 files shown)


app/controllers/subscriptions_management_controller.rb
```diff
@@ -101,6 +102,13 @@ def use_govuk_account_layout?
 
 private
 
+  def handle_one_login_hint
+    return unless params[:from] == "your-services" && !authenticated?
+
+    redirect_path = list_subscriptions_path
```
1pretz1, line 108: Nitpick: probably don't need the variable due to it only being used once

spec/controllers/subscriptions_management_controller_spec.rb
```diff
@@ -40,6 +41,32 @@
...
+        stub_account_api_get_sign_in_url(
+          redirect_path: "/email/manage",
+          auth_uri:,
+        )
+      end
+
+      let(:auth_uri) { "/sign-in" }
+
+      context "with a login session" do
+        it "ignores the " do
```
1pretz1, line 55: Missing the rest of the test name?

1pretz1 (approved):
Looks good! Just some small, non-blocking comments 🍏

app/controllers/subscriptions_management_controller.rb
KludgeKML: Nice - moved into the next line.

PR #1630, opened 2023-10-09
Simplify One Login interstitial page

KludgeKML:
Simplify the interstitial page that comes when a user comes via the single-page-notification-button path and is required to create/use a One Login account.

https://trello.com/c/yxBLErqj/2135-improve-account-journeys-in-email-alert-frontend

BEFORE:
Screenshot 2023-10-09 at 11 04 16

AFTER:

Screenshot 2023-10-09 at 12 13 25

### DIFF
--- a/app/views/govuk_account_signups/show.html.erb
+++ b/app/views/govuk_account_signups/show.html.erb
@@ -11,36 +11,9 @@
   <%= t("single_page_subscriptions.account_required_description") %>
 </p>
 
-<%= render "govuk_publishing_components/components/heading", {
-  text: t("single_page_subscriptions.subheading"),
-  heading_level: 2,
-  font_size: "m",
-  margin_bottom: 6
-} %>
-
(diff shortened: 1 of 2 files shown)


jon-kirwan (commented):
Mostly looks fine to me but I have one comment. Should we include HTML in YAML? Perhaps we should move the `<p class="govuk-body">...</p>` to the view and only keep the text in the YAML file?

KludgeKML:
@jon-kirwan 

> Mostly looks fine to me but I have one comment. Should we include HTML in YAML? Perhaps we should move the `<p class="govuk-body">...</p>` to the view and only keep the text in the YAML file?

Oh, yes definitely. I suspect that that paragraph once upon a time had something complicated in it that meant it could only be HTML, but it definitely doesn't need to have it now.

matthillco (approved):
With Jon's change, it looks good to me.

PR #1812, opened 2024-07-04
Remove Universal Analytics 

AshGDS:
## What / Why
- Removes Universal Analytics now that GA4 is the current analytics offering
- https://trello.com/c/fKGle0gD/187-remove-ua-code-from-applications, [Jira issue PNP-8334](https://gov-uk.atlassian.net/browse/PNP-8334)

## Note
There's something called `redirect_with_analytics` in this repo but it seems backend related, so maybe that's something for a backend dev to sort out? I think it adds cross domain tracking to redirects that comes from `govuk_personalisation` (see [here](https://github.com/alphagov/govuk_personalisation/blob/ac46fe2a88b5937ba5d12b12c7e0128bafd57a21/lib/govuk_personalisation/controller_concern.rb#L138)).


andysellick (approved):
👍

